PR #48, opened 2021-03-02
relax constraints on Send in System and SpawnBuilder

mcginty:
On integrating the new actor system with tonari's main codebase, the type system showed more restrictiveness than was necessary, making it impossible to run an `Actor` on the main thread if `Actor` didn't implement `Send` (which of course is unnecessary if not being spawned on a separate thread).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -193,7 +193,7 @@ pub struct SpawnBuilder<'a, A: Actor, F: FnOnce() -> A> {
     system: &'a mut System,
     capacity: Option<usize>,
     addr: Option<Addr<A>>,
-    factory: Box<F>,
+    factory: F,
 }
 
 impl<'a, A: 'static + Actor, F: FnOnce() -> A> SpawnBuilder<'a, A, F> {
@@ -265,7 +265,7 @@ impl System {
         A: Actor + 'static,
         F: FnOnce() -> A + Send + 'static,
     {
-        SpawnBuilder { system: self, capacity: None, addr: None, factory: Box::new(factory) }
+        SpawnBuilder { system: self, capacity: None, addr: None, factory }
     }
 
     /// Spawn a normal [`Actor`] in the system, returning its address when successful.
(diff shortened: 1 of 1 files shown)


bschwind:
Could we get a minimal example added which would give a compiler error before this fix is added?

mcginty:
Added a test cc @bschwind

mcginty:
It's not easy to show in this PR, but if you copy that test and run it on `main` it doesn't compile because of `Send` constraints.

bschwind (approved):
Nice change! The test you added is really helpful to understand what should be allowed, thanks for that.

skywhale (approved):
LGTM

strohel:
I'd like to have a look at this later today, unless you want to merge quickly.

src/lib.rs
```diff
@@ -258,9 +260,9 @@ impl System {
     /// created on its own thread instead of the calling thread.
     /// Returns a [`SpawnBuilder`] which can be used to customize the
     /// spawning of the actor.
-    pub fn prepare_fn<A, F>(&mut self, factory: F) -> SpawnBuilder<A>
+    pub fn prepare_fn<A, F>(&mut self, factory: F) -> SpawnBuilder<A, impl FnOnce() -> A>
```
skywhale, line 263: This is more explicit, I think.
```suggestion
    pub fn prepare_fn<A, F>(&mut self, factory: F) -> SpawnBuilder<A, F>
```

mcginty: Oh yeah totally.

src/lib.rs
```diff
@@ -189,14 +189,14 @@ pub struct Context<A: Actor + ?Sized> {
 /// whether to spawn the Actor into its own thread or block
 /// on the current calling thread.
 #[must_use = "You must call .spawn() or .block_on() to run this actor"]
-pub struct SpawnBuilder<'a, A: Actor> {
+pub struct SpawnBuilder<'a, A: Actor, F: FnOnce() -> A> {
     system: &'a mut System,
     capacity: Option<usize>,
     addr: Option<Addr<A>>,
-    factory: Box<dyn FnOnce() -> A + Send + 'static>,
+    factory: Box<F>,
```
strohel, line 196:
```suggestion
    factory: F,
```
I was unsure whether FnOnce() is Sized or not, but apparently yes, this seems to work for me.

strohel (commented):
Looks good, it seems we can even remove the Box.

PR #49, opened 2021-03-03
chore: remove one boxing

strohel:
Kinda interesting that this worked before. Apparently, Box<Fn*> implements Fn*:
https://doc.rust-lang.org/std/boxed/struct.Box.html#impl-FnOnce%3CArgs%3E


mcginty:
Yay thanks! Sorry I missed that one.

Issue #14, opened 2021-02-19
Built-in timers/intervals


mcginty:
Do we want??

bschwind:
Some resources that will be helpful here:

https://github.com/tokio-rs/tokio/tree/master/tokio-util/src/time

https://crates.io/crates/hierarchical_hash_wheel_timer

bschwind:
A simple, naive implementation has been implemented in #52 but I'd like to eventually replace that implementation with #53 after settling on a user-facing API.

strohel:
#53 has been merged, but per comments therein we still want to introduce a helper actor + an ergonomic API on top of it. I'll open a PR to iterate that.

PR #51, opened 2021-04-28
Timer

bschwind:
From #14

It will be nice to be able to schedule timers and one-off tasks for periodically sending messages to actors, instead of having the actor send to itself and sleep its thread for certain intervals.


bschwind:
Closing in favor of #52

PR #52, opened 2021-05-06
Simple Timer

bschwind:
From #14

This feels a lot simpler than using `hierarchical_hash_wheel_timer`.

It does the dumb thing which is to just keep a `Vec` of tasks and tick every millisecond (or a configurable timer resolution). It iterates over each task and runs if it has expired, and reschedules it if it's a recurring task.

CPU usage is noticeably lower compared to `hierarchical_hash_wheel_timer` but it likely doesn't scale as well with lots and lots of timers in comparison.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -41,7 +41,7 @@
 //! ```
 //!
 
-use crate::timer::{ScheduleToken, TimerControlFlow, TimerHandle, TimerRef};
+use crate::timer::{TimerHandle, TimerRef};
 use crossbeam_channel::{self as channel, select, Receiver, Sender};
 use log::*;
 use parking_lot::{Mutex, RwLock};
@@ -50,7 +50,9 @@ use std::{fmt, ops::Deref, sync::Arc, thread, time::Duration};
 #[cfg(test)]
 pub mod testing;
 
-pub mod timer;
+mod timer;
+
+pub use timer::ScheduleToken;
 
 // Default capacity for channels unless overridden by `.with_capacity()`.
 static DEFAULT_CHANNEL_CAPACITY: usize = 5;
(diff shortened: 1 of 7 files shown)


src/timer.rs
```diff
@@ -0,0 +1,196 @@
...
+                    match entry.timer_type {
+                        TimerType::OneShot(closure) => {
+                            closure(schedule_token);
+                        },
+                        TimerType::Recurring(interval, mut closure) => {
+                            match closure(schedule_token) {
+                                TimerControlFlow::Continue => {
+                                    let new_entry = TimerEntry {
+                                        schedule_token,
+                                        scheduled_at: entry.scheduled_at + interval,
```
mcginty, line 180: Do we need to add code to check that the runtime wasn't so long that this `scheduled_at` would actually be in the past?

src/timer.rs
```diff
@@ -0,0 +1,196 @@
...
+    timer_ref: TimerRef,
+}
+
+#[derive(Debug, Clone)]
+pub struct TimerRef {
+    thread_tx: Sender<ThreadMessage>,
+}
+
+impl TimerRef {
+    pub fn run_once<F>(&mut self, delay: Duration, callback: F) -> ScheduleToken
```
mcginty, line 56: could this just call `run_recurring` internally with a wrapping closure that sends `TimerControlFlow::Cancel` or is that too convoluted? that way we wouldn't need `OneShot` and `Recurring`

bschwind: I'd prefer to keep it as-is, it feels nicer to use `FnOnce` for the "send once" case and `FnMut` just for the "send recurring" case.

src/timer.rs
bschwind: If we cared a bit more about timer precision then yes, but in the case you describe, it should be picked up and run in the next loop iteration.

We could definitely make improvements to measure elapsed time after doing the task work and sleep less in that case, but that veers from the intended simplicity I had for this timer.

src/lib.rs
```diff
@@ -181,6 +186,89 @@ pub struct SystemHandle {
...
     pub myself: Addr<A>,
+    timer: TimerRef,
+}
+
+impl<A: Actor + ?Sized> Context<A> {
+    /// Sends a message once to the calling actor itself, after a delay.
+    /// Cancel with a call to `context.cancel_timer(schedule_token);`
+    pub fn send_once_to_self(&mut self, delay: Duration, msg: A::Message) -> ScheduleToken {
+        self.send_once_to::<A>(delay, msg, self.myself.recipient.clone())
+    }
```
strohel, line 197: Public API wise, it feels somewhat unsymmetrical that `send()` is on `Recipient`, while `send_{once,recurring}*()` are on `Context`.
Do I get it right that it is because we didn't want to stick another field (`TimerRef`) to `Recipient`?

strohel (commented):
Just one quick question so that I understand better some design choices.

I'm also playing with an idea in my head that both "send delayed" and "send recurring" could be implemented using forms of "wrapping relay actors" (with pros and cons). What would be problematic is cancellation and not blocking that actor's shutdown (control) channel for too long. That would probably call for some "interruptible sleep in an actor". Let's see if this idea survives sleeping on it.

src/lib.rs
skywhale: Agree that `context.myself.send_once(...)` reads more symmetric. It allows calling the function outside the actor, too, whether it's a good idea or not. `TimerRef` seems lightweight enough to copy to each `Recipient`?

bschwind: I'll try this out! I think it makes sense and it's not super heavy on the recipient.

src/lib.rs
```diff
@@ -230,19 +318,42 @@ impl<'a, A: 'static + Actor, F: FnOnce() -> A + Send + 'static> SpawnBuilder<'a,
     }
 }
 
+fn run_once<F>(timer: &mut TimerRef, delay: Duration, callback: F) -> ScheduleToken
```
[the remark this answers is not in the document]
bschwind: Good call!

src/lib.rs
bschwind: So the immediate issue I run into here is: How do I construct an `Addr` or `Recipient` without involvement of the actor `System`?

Right now users of the crate can do something like:

```rust
let addr = Addr::default();
let another_addr = Addr::with_capacity(50);

let recipient: Recipient<()> = addr.recipient();
```

However there is no way to access a reference to the timer thread without passing it explicitly, which I think is a major hit on ergonomics. Am I missing anything?

skywhale: Yeah, we'll need to make it `System::addr()` and `System::addr_with_capacity()` I suppose.

strohel:
As a mental exercise, I have submitted #53 which could be (if finished) an alternative approach. It has both pros and cons. But I don't know whether there are some specific requirements, or perhaps some other work is depending on this - I'm cool with that experiment ending up being just ..an experiment. :)

bschwind:
@strohel I'd like to merge and release this PR for now, and have #53 be a replacement once the final design for it is settled. It'll be an extra semver bump but no big deal I suppose. I've already done quite a bit of testing for this PR on tonari's codebase and unfortunately I don't have a lot of time to do more for #53 yet.

strohel:
@bschwind I see, cool. In fact, merging this removes any possible time pressure from #53, which is good.

With that in mind, the longer iteration time I introduce working on that sporadically should no longer matter, so let me try to iterate #53 further (unless you want to yourself, of course).

bschwind:
@strohel Ah yep, definitely no time pressure to get #53 working! I think overall you have some better ideas swirling around in your head for the implementation and I'm excited to see the final result of that. It'll feel less gross than this change for sure. We just had a fairly pressing need to remove all the `thread::sleep()` calls we were using previously to schedule messages. `tonari-actor` is now officially being used by tonari though, so yay to that 🎉

PR #53, opened 2021-05-10
Implement message receive timeout, the Actor::deadline_passed() callback

strohel:
This is a result of an experimentation triggered by #52, and relates itself to #14.

Direction-wise this is an alternative to #52, but this PR does not implement a full-fledged timer, merely a building block on top of which one can be made. Simplified timer is presented as an example.

The main differentiating point is a focus on minimal additions to the core actor crate by allowing downstreams to implement their own in form of an Actor (though we can ship some ourselves). The design should allow implementing timers without any permanent "ticks", with resolution of [`thread::park_timeout()`](https://doc.rust-lang.org/std/thread/fn.park_timeout.html) used by crossbeam-channel and ability to be interrupted by an incoming message.

From the API user PoV, the difference is that with this approach, the user has the burden of carrying some `Recipient<TimerActor>` around, while in #52, `tonari-actor` itself has to stick `TimerRef` somewhere.

The first 2 commits are preparatory ones (first extracted from Brian's PR), `Implement Context::set_deadline() and Actor::deadline_passed() callback` is the main one.


bschwind (commented):
This looks great! I like the internal implementation much more than #52 - it fits more cleanly with crossbeam-channel, and you don't have to pay for things you don't use (timer thread, 1ms tick resolution, etc.)

As you mentioned, the user API is less ergonomic than I'd like and that could be remedied with some built-in actors we provide (which don't have any runtime cost if not used).

I think the main use case we want to target is a multitude of actors which mostly send messages to themselves at a fixed interval (at least, that's what we end up needing the most in tonari :P ) The part I want to avoid is having to collect some `Vec` of actor addresses together and pass that to the timer actor. It would be good if setting up a timer interval can happen entirely in the actor's lifecycle callbacks, such as in `started` or `handle`.

Let me know if you'd like me to make an attempt at adding an ergonomic user API on top of this, or if you already have some ideas swirling around in your head.

Overall I think this is great though and I'd much prefer it to #52 if we can get the API right.

skywhale:
I didn't realize that `crossbeam-channel` provides timer functionalities. Good to know :)

@bschwind, as far as I know, all the actors who send messages themselves handle only one fixed-interval cycle. In that case, using @strohel's framework, would the following be sufficient?

```
impl Actor for SomeActor {
  fn deadline_passed(...) {
    context.myself.send(Message::Check);
    context.set_deadline(CHECK_INTERVAL);
  }
  fn handle(...) {
    match message {
      Message::Check => {
        do_some_recurring_stuff();
      },
    }
  }
}
```

or even:

```
impl Actor for SomeActor {
  fn deadline_passed(...) {
    do_some_recurring_stuff()
    context.set_deadline(CHECK_INTERVAL);
  }
}
```

When we have a more complex actors who have multiple recurring cycles at different intervals, then we can use something along the line of `DelayActor` example

bschwind:
@strohel thanks for your research on this! I'll try integrating this branch into tonari along with @skywhale 's code suggestion above. If it works well we can try fixing up the API a bit more to include a convenience wrapping actor, and/or a more ergonomic API on top of it all.

skywhale:
I think (A) makes the most sense, if we can change the semantics of `deadline_passed` a bit, so it's not reset by incoming messages. Otherwise, I think the building block is a little too tricky to use. I like the flexibility of the original `set_deadline` and `deadline_passed`, assuming the changed semantics. Not to mention `TickingWrapper`, users can easily implement various types of timing logic.

(A) not as accurate as `TickingWrapper`, but something as simple as this can implement approximate ticks.
```
fn deadline_passed() {
  context.set_deadline(INTERVAL);
  do_recurring_stuff();
}
```

(B) if `do_recurring_stuff()` is an expensive call that take a few milliseconds, this pattern may be more appropriate.

```
fn deadline_passed() {
  do_recurring_stuff();
  context.set_deadline(INTERVAL);
}
```

strohel:
@bschwind, @skywhale, thank you for your feedback. It made me understand that setting (rather than querying) the deadline could indeed be a more ergonomic API, and that we can preserve the deadline even when a message is received.

I've force-pushed a new version that incorporates both ideas and is rebased on `main`. The first 3 commits are preparatory and partially revert the previous implementation, the fourth one is the actual functionality. So the overall PR diff shows the difference between the 2 implementations and usage in the timer example, and the "pure additions" can be seen by looking at the last commit in isolation.

examples/delay_actor.rs
```diff
@@ -0,0 +1,145 @@
...
+use std::{
+    cmp::Ordering,
+    collections::BinaryHeap,
+    time::{Duration, Instant},
+};
+use tonari_actor::{Actor, Context, Recipient, SendError, System};
+
+/// Something that can be sent. A message with a recipient.
+trait Envelope {
+    fn send(self: Box<Self>) -> Result<(), SendError>;
```
strohel, line 13: Trivia: it was tricky for me to figure out the need to wrap `Self` in a `Box` here. But without it, calling `message.send()` given `message: Box<dyn Envelope + Send>` didn't work: `rustc` complained it cannot move a type whose size is not known at compile time. Wrapping in a Box moves the move (heh) into the trait implementation, where it is known. [Method call operator still works](https://doc.rust-lang.org/reference/items/associated-items.html#methods).

examples/simple_timer.rs
```diff
@@ -43,6 +38,19 @@ impl Actor for TimerExampleActor {
         println!("Got a message: {:?} at {:?}", message, self.started_at.elapsed());
         Ok(())
     }
+
+    fn deadline_passed(&mut self, context: &mut Context<Self>, _deadline: Instant) {
+        context.myself.send(TimerMessage::Periodic).unwrap();
+
+        // A: Schedule one second from now (even if delayed); drifting tick.
```
skywhale, line 45: Could you please move this documentation and examples to `lib.rs`? I think it's crucial to understand how to use the `deadline` effectively.

src/lib.rs
```diff
@@ -182,93 +183,16 @@ pub struct SystemHandle {
...
-    /// Do not rely on precise timing of the cancellation, the message _may_ be sent
-    /// before it is cancelled.
-    pub fn cancel_timer(&mut self, schedule_token: ScheduleToken) {
-        self.timer.cancel(schedule_token);
-    }
-
-    pub fn shutdown(&mut self) -> Result<(), ActorError> {
-        self.system_handle.shutdown(&mut self.timer)
-    }
+    /// A deadline for receiving the next message.
```
skywhale, line 190: My old C++ self tells me that `receive_deadline` should be private, and there should be `set_deadline()`. Is `Rust` not opinionated about it?

src/lib.rs
```diff
@@ -692,6 +616,12 @@ pub trait Actor {
 
     /// An optional callback when the Actor has been stopped.
     fn stopped(&mut self, _context: &mut Context<Self>) {}
+
+    /// An optional callback when deadline has passed while waiting for a message.
```
skywhale, line 620: nit: we can remove `while waiting for a message` now.

src/lib.rs
```diff
@@ -934,4 +867,68 @@ mod tests {
...
+        addr.send(Some(Instant::now() - Duration::from_secs(1))).unwrap();
+        thread::sleep(Duration::from_millis(10));
+        assert_eq!(handle_count.load(Ordering::SeqCst), 1);
+        assert_eq!(timeout_count.load(Ordering::SeqCst), 1);
+
+        // Test the normal case.
+        addr.send(Some(Instant::now() + Duration::from_millis(20))).unwrap();
+        thread::sleep(Duration::from_millis(10));
+        assert_eq!(handle_count.load(Ordering::SeqCst), 2);
+        assert_eq!(timeout_count.load(Ordering::SeqCst), 1);
```
skywhale, line 917: This test can be flaky. Does Rust testing framework support marking such tests?

skywhale (approved):
Thank you @strohel :)

As the next step, it's helpful to land https://github.com/tonarino/actor/pull/54 and provide an actor wrapper that implements a similar API to @bschwind's `send_recurring_*` functions. Assuming it's a popular use case of the actor system, we can put it in a feature-gated `util` mod instead of an example.

src/lib.rs
strohel: I was hesitant whether to make it pub or use getters/setters. Rust resources aren't unanimous on this:
- https://rust-lang.github.io/api-guidelines/future-proofing.html#c-struct-private warns that pub fields are a stronger commitment, and goes on to say:
  > Public fields are most appropriate for struct types in the C spirit: compound, passive data structures. Otherwise, consider providing getter/setter methods and hiding fields instead.
- https://www.reddit.com/r/rust/comments/d7w6n7/is_it_idiomatic_to_write_setters_and_getters/f15ib88?utm_source=share&utm_medium=web2x&context=3 leans on the "pub fields" side, noting rust specifics to destructuring, but is somewhat opinionated.
- https://users.rust-lang.org/t/public-getter-method-vs-pub-field/20147 contains very informative unopinionated summary of all implications, highlighting construction and partial borrowing. I liked matklad's (the author of rust-analyzer) rule of thumb there:
  > - If there’s a combination of values of fields which breaks some invariant, than make all fields private and provide a getter
  > - If any combination of fields values makes sense, make all fields public
- there were even bug reports to rust projects to make private fields `pub`, addressed: https://github.com/rust-syndication/rss/issues/71

In our case, all permissible values of `receive_deadline: Option<Instant>` are valid at any point of time, and it doesn't interact with other fields. That, and simplicity probably made me lean to the "field side".

Adding a private field would have a side-effect: it would make whole `Context` non-constructible by external crates unless we provided a `pub new(...) method. That would interact with actor wrapping unless we added some `pub wrap(...) method. Independently of PR, I actually think that might be a good idea, but it would probably fit better as a separate PR (and if done using `#[non_exhaustive]` we could keep the fields public if we wish so).

src/lib.rs
strohel: > This test can be flaky.

The intention is that they are not flaky in practice - that we sleep at least 10 ms between writing a value in one thread and asserting on it in another thread. I didn't hit a failed test even if I lowered the delay to 1 ms (with 0 ms wait I did). That isn't of course a proof that it won't be flaky.

Perhaps I've overlooked something? My naive assumption was that a failure can occur only when OS scheduler is very unfair and schedules thread A _after_ thread B that requested to be woken 10 ms before A requested to be woken.

> Does Rust testing framework support marking such tests?

I have only found an `#[ignore]` attribute, which is rather coarse https://doc.rust-lang.org/reference/attributes/testing.html#the-ignore-attribute
More loosely related, I had nice experience with parametrized testing using https://github.com/frondeus/test-case

src/lib.rs
bschwind: My gut feeling is that we want this private, with some convenience functions for setting a specific deadline (`context.set_deadline(Instant)`) or timing out after some duration (`context.expire_deadline_after(Duration)`)

src/lib.rs
```diff
@@ -406,6 +434,11 @@ impl System {
                         }
                     }
                 },
+
+                recv(timeout) -> _ => {
+                    let deadline = context.receive_deadline.take().expect("implied by timeout");
+                    actor.deadline_passed(context, deadline);
```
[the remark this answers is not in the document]
bschwind: What do you think about having `deadline_passed()` return either a custom enum or an `Option<Instant>` which we use to assign the `receive_deadline`? I think the burden is a bit too high to have to remember to set the field back to `None` - probably better to have the type system force you to think about it explicitly.

bschwind (commented):
Looking good so far! I have a few questions on how the public interface should be structured but I'm liking the overall change here. Burns a lot less CPU too :)

Is the assumption right now that we'll add in some built-in convenience actors (perhaps under a feature flag as @skywhale mentioned) in a separate followup PR?

src/lib.rs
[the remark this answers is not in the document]
strohel: > I think the burden is a bit too high to have to remember to set the field back to None - probably better to have the type system force you to think about it explicitly.

Not sure if I get you right here - `receive_deadline` _is_ currently reset to `None` right before calling `deadline_passed()` - by the `context.receive_deadline.take()`.

> What do you think about having deadline_passed() return either a custom enum or an Option<Instant> which we use to assign the receive_deadline?

I was considering that, but we'd then have 2 competing ways to set the deadline - which could get confusing for example when `deadline_passed()` delegates the deadline setting to some other function. (we cannot really remove the setters, deadline can also be set from hande(), started() etc.)

src/lib.rs
strohel: Alright, I've changed this to private and introduced `deadline()`, `set_deadline()` and `set_timeout()` methods in a just-pushed version. I'll deal with the wrap-ability of `Context` in #54 later.

strohel:
> As the next step, it's helpful to land #54 and provide an actor wrapper that implements a similar API to @bschwind's send_recurring_* functions. Assuming it's a popular use case of the actor system, we can put it in a feature-gated util mod instead of an example.

> Is the assumption right now that we'll add in some built-in convenience actors (perhaps under a feature flag as @skywhale mentioned) in a separate followup PR?

Right, I count with opening a follow-up PR to extend and promote `DelayActor`, possibly after landing #54 as Ryo suggested.

I've pushed new commit that should address most review points above (thanks for them). I've resolved relevant conversations. (do you have a preferred convention who resolves/when to resolve them?)

src/lib.rs
skywhale: Thank you for all the discussion pointers @strohel. To `matklad`'s point, `Context` is not a passive data container, and it's easy to see that it may evolve in the future to break the invariance. As you suggested, I think it makes sense in a follow-up PR to make all the fields private. It should by by design that the clients are not able to construct `Context`, and it's only given by the `actor` system.

src/lib.rs
skywhale: You didn't overlook anything. I was referring to the case when the OS scheduler is very unfair e.g. <1% of time on a significantly overloaded CI machine :) Thanks for testing with 1ms granularity and also checking the test passes. It's good as is, and this conversation is a useful reference for later :)

bschwind (approved):
I don't have any further comments, seems solid to me!

Regarding resolving conversations, you can feel free to resolve ones you've addressed via code changes, or if the conversation seems to have concluded (maybe confirm with the person who started the original thread). We don't have any strong rules for this either way.

skywhale (approved):
LGTM++

PR #54, opened 2021-05-12
Context: change myself field from Addr<A> to Recipient<M>

strohel:
This makes whole Context parametrised by M (message), rather than A: Actor.
As a result, Context is more flexible to pass on, which facilitates creation
of Actor wrappers.

This changes the signature of the Actor trait methods, which breaks the API,
though required edits should be mostly mechanical. The field `myself` was
arguably used most for sending messages to oneself, which already went
through a `Deref` to `Recipient`.

One piece of functionality was lost on Context: the `Addr::recipient()`
method which can produce a "converting" recipient.
`recipient()` can be moved from `Addr` to `Recipient` to preserve it.
Do you think it is needed?

### DIFF
--- a/benches/benches.rs
+++ b/benches/benches.rs
@@ -16,7 +16,7 @@ impl Actor for ChainLink {
 
     fn handle(
         &mut self,
-        context: &mut Context<Self>,
+        context: &mut Context<Self::Message>,
         message: Self::Message,
     ) -> Result<(), Self::Error> {
         if message > 0 {
--- a/examples/actor_wrapping.rs
+++ b/examples/actor_wrapping.rs
@@ -1,0 +1,78 @@
+use anyhow::Error;
+use env_logger::Env;
+use log::debug;
+use std::time::{Duration, Instant};
+use tonari_actor::{Actor, Context, System};
+
+/// An actor that wraps any other actor and adds some debug logging around its calls.
+struct LoggingAdapter<A> {
+    inner: A,
+}
+
+impl<A: Actor> Actor for LoggingAdapter<A> {
+    type Error = A::Error;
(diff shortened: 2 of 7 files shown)


skywhale (approved):
Thank you @strohel :)

Semantically `Context` parameterized by `Actor` makes more sense to me, but I see that being able to wrap an actor provides a useful framework for sharing specialized lifecycle methods across multiple actors in the client code. It should be okay to drop `Context::myself::recipient()` functionality.

Please wait for @bschwind's review, too. As he's got more context on the `actor` crate than I.

strohel:
Note that this is no longer technically a prerequisite of #53 as it no longer depends on wrapping actors. But wrapping actors in general could be still useful, and so could be this PR. Downside is the scale of edits that downstreams would need to make to accommodate this.

There may also be alternative technical approaches. Ability to create `Context<BarActor>` from `&mut Context<FooActor>` (perhaps when `FooActor::Message` implements `Into/From<BarActor::Message>`) would be one such option. But that approach also looks like requiring a change to the API (and the internals). And `message_rx: Receiver<A::Message>`, which is currently transitive member of `Context<A>`, may prevent us from providing such a wrapper. The same `message_rx` is being removed from `Context` by this PR.

bschwind (approved):
This seems totally reasonable to me! We weren't using anything special from the `Actor` implementations themselves, just their associated `Message`. If we can't think of anything particular we'd want to use from the `Actor` itself, I'm cool with merging this.

strohel:
I've rebased this on master and added an actor wrapping example. We've got quite o lot of examples now, but I'll move delay_actor out soon. I'll merge this tomorrow unless something pops up.

PR #57, opened 2021-05-26
Add an option for unbounded actor addresses

bschwind:
In some cases we create actors with "large enough" buffers, like `500` or so. In these cases, instead of just guessing the capacity, it probably makes more sense to just use an unbounded channel. The `500` was essentially saying "we should have enough where it never reaches capacity", and I think using `unbounded` makes that distinction more clear.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -216,6 +216,11 @@ impl<M> Context<M> {
     }
 }
 
+enum ChannelCapacity {
+    Bounded(usize),
+    Unbounded,
+}
+
 /// A builder for specifying how to spawn an [`Actor`].
 /// You can specify your own [`Addr`] for the Actor,
 /// the capacity of the Actor's inbox, and you can specify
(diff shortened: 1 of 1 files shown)


mcginty:
I'm kind of hesistant to support unbounded actor addresses, since it opens up the possibility for unbounded memory consumption and might lead to issues that are more difficult to track down compared to errors that can be caught within the binary.

bschwind:
I don't share those same concerns. In the same way that `crossbeam-channel` provides `unbounded` and it's up to you to use it responsibly, why can't this crate provide the same?

The problem we want to solve is "This actor is not on the 'hot path' of my application. I want it to receive every message it gets sent, even if it takes a little longer or uses up some memory."

That's why we have some actors with an addr of `Addr::with_capacity(500)`. It's kind of silly to just take a random guess at what that value should be if your intention is for it to never drop messages. I'm talking about actors which handle state/config changes which don't happen at a regular frequency but which may have some burstiness to it.

strohel (commented):
I initially thought this is fine, but after some thinking added a https://github.com/tonarino/portal/pull/981/files#r639633017 comment exploring alternatives to having to introduce unbounded channels now.

I don't dare to say there won't be any valid use-case for unbounded channels in actor in the future. But maybe we don't have a valid use-case _yet_, and we can safely postpone this until we're more informed.

skywhale:
@bschwind I'm closing this PR for now.

PR #58, opened 2021-07-13
Add Timed actor wrapper + friends for delayed/recurring messages

### LINKED ISSUE #14: Built-in timers/intervals

strohel:
See module level docs and `delay_actor.rs` example for info.

When actor A wants to send a delayed message to actor B, there are 3 places where the queue can be:
1. in actor (thread) A
2. in some separate actor (thread) C
3. in actor (thread) B.

This PR choses option 3. In our downstream usage, all recurring messages are "to self", so options 1. and 3. are functionally (but not API-wise) equivalent. The chosen approach has some pros and cons:
- pro: modulo the preparatory "independent" core API changes, Timed<> wrapper and friends are strictly separate form the core API (could have been implemented in a separate crate).
- pro: avoids having to pass some `DelayActor` reference around, or having it registered as some global resource.
- pro: delayed messages do not cross thread boundary (in our use-case).
- big con: requires wrapping each delayed-message-receiving actor in the wrapper.
- con: requires ability to customize the `Context` type in actor as a prerequisite.

Closes #14.

Separated into atomic commits, reviewing them individually is suggested. The last commit adds the main functionality, others are prerequisites. I had to tweak core actor API a bit (with changes needing to be done downstream) to be able to provide a nice API of the delayed/recurring functionality.

Demonstration of downstream code changes tonarino/portal#1126, specifically https://github.com/tonarino/portal/pull/1126/commits/4cfefdd45ba2ff7155c6e1943ba90d1bc20c6fe0.

### DIFF
--- a/benches/benches.rs
+++ b/benches/benches.rs
@@ -17,7 +17,7 @@ impl Actor for ChainLink {
 
     fn handle(
         &mut self,
-        context: &mut Context<Self::Message>,
+        context: &mut Self::Context,
         message: Self::Message,
     ) -> Result<(), Self::Error> {
         if message > 0 {
--- a/benches/benches.rs
+++ b/benches/benches.rs
@@ -7,6 +7,7 @@ struct ChainLink {
 }
 
 impl Actor for ChainLink {
+    type Context = Context<Self::Message>;
     type Error = ();
     type Message = u64;
 
--- a/src/timed.rs
+++ b/src/timed.rs
@@ -1,0 +1,220 @@
+//! Tools to make a given actor able to receive delayed and recurring messages.
+//!
+//! To apply this to a given (receiving) actor:
+//! * Use [`TimedContext<Self::Message>`] as [`Actor::Context`] associated type.
+//!   * Such actors cannot be spawned unless wrapped, making it impossible to forget wrapping it.
+//! * Wrap the actor in [`Timed`] before spawning.
+//!
+//! The wrapped actor will accept [`TimedMessage<M>`] with convenience conversion from `M`.
+//! [`RecipientExt`] becomes available for [`Recipient<TimedMessage<M>>`]s which provides methods like
+//! `send_delayed()`, `send_recurring()`.
+//!
+//! See `delay_actor.rs` example for usage.
+
+use crate::{Actor, Context, Recipient, SendError, SystemHandle};
+use std::{
+    cmp::Ordering,
+    collections::BinaryHeap,
+    ops::Deref,
+    time::{Duration, Instant},
+};
+
+/// A message that can be delivered now, at certain time and optionally repeatedly.
+pub enum TimedMessage<M> {
+    Instant { message: M },
+    Delayed { message: M, fire_at: Instant },
+    Recurring { factory: Box<dyn FnMut() -> M + Send>, fire_at: Instant, interval: Duration },
+}
+
+/// This implementation allows sending direct unwrapped messages to wrapped actors.
+impl<M> From<M> for TimedMessage<M> {
+    fn from(message: M) -> Self {
+        Self::Instant { message }
+    }
+}
+
+/// Convenience methods for [`Recipient`]s that accept [`TimedMessage`]s.
+pub trait RecipientExt<M> {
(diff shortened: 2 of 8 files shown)


skywhale (commented):
I only had a brief look, but I really liked how you implemented `timed` as an optional capability on top of the existing actor API. I will do one more pass tonight.

src/timed.rs
```diff
@@ -0,0 +1,220 @@
...
+
+    fn deadline_passed(
+        &mut self,
+        context: &mut Self::Context,
+        _deadline: Instant,
+    ) -> Result<(), Self::Error> {
+        // Handle all messages that should have been handled by now.
+        let now = Instant::now();
+        while self.queue.peek().map(|m| m.fire_at <= now).unwrap_or(false) {
+            let item = self.queue.pop().expect("heap is non-empty, we have just peeked");
```
mcginty, line 157: Not a comment on your code, but it'd be nice if they also had a [`drain_filter()`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.drain_filter) function for `BinaryHeap`.

benches/benches.rs
```diff
@@ -16,7 +17,7 @@ impl Actor for ChainLink {
 
     fn handle(
         &mut self,
-        context: &mut Context<Self::Message>,
```
mcginty, line 19: I like what you did with the context simplification!

mcginty (approved):

skywhale (approved):
This looks great! I couldn't find anything to comment on :)

src/timed.rs
```diff
@@ -0,0 +1,220 @@
...
+
+use crate::{Actor, Context, Recipient, SendError, SystemHandle};
+use std::{
+    cmp::Ordering,
+    collections::BinaryHeap,
+    ops::Deref,
+    time::{Duration, Instant},
+};
+
+/// A message that can ne delivered now, at certain time and optionally repeatedly.
```
bschwind, line 22: ne -> be

src/timed.rs
```diff
@@ -0,0 +1,220 @@
...
+
+/// Implementation detail, element of message queue ordered by time to fire at.
+struct QueueItem<M> {
+    fire_at: Instant,
+    payload: Payload<M>,
+}
+
+impl<M> PartialEq for QueueItem<M> {
+    fn eq(&self, other: &Self) -> bool {
+        self.fire_at == other.fire_at
```
bschwind, line 197: Is it okay that the `payload` field isn't considered here? If you scheduled two messages to fire at the same by reusing some `Instant`, would they both fire?

bschwind (commented):
This is looking great! Really impressive type-level hackery going on. I'm mostly done reading it but still need another sitting to fully internalize how it all works.

src/timed.rs
strohel: It would! I don't like using `expect()` for something that could be proved infallible at compile time.

benches/benches.rs
strohel: Well OTOH I introduce the responsibility for each actor to define the `type Context = ...`. I wish [associated type defaults](https://github.com/rust-lang/rust/issues/29661) were stable, but it doesn't look even close.

src/timed.rs
strohel: Good :eyes:! Fixed.

src/timed.rs
strohel: That should be safe: we never compare queue items for equality ourselves and we never deduplicate them any any way. `PartialEq` and friends are just a requirement of `BinaryHeap`.

I wish there existed some Python-style `BinaryHeap::with_key_func<K: Ord>(f: Fn(V) -> K)` so that we can define the comparison only in the context of the container and don't mess with global `{Partial,}{Eq,Ord}`.

strohel:
Thanks for the reviews!

I'll let this simmer down for a couple of days (if @bschwind or @skywhale want to give this another read), then I'd like to merge and cut a 0.6.0 release

bschwind (approved):
I took another pass through it, I think it's looking great and ready to merge!

PR #59, opened 2021-07-26
RFC: Include intended recipient name in SendError

strohel:
#### Add test for SendError

We're going to touch it, the tests will demonstrate the changes.

### Include intended recipient name in SendError

This aids with debugging send errors. Thanks to type decoupling made by
`Recipient`, senders may not even know the actual receiving actor, so
senders cannot replicate this behaviour fully on their own.

Should help with debugging downstream issues like tonarino/portal#1153.

Caveats:
- Increases the size of `Recipient`, `Addr`, `SendError` by ~one pointer~ two pointers (one `&str`).
- Introduces API-breaking changes to `SendError` and `SendResultExt::on_full()`.
  - ~Specifically, SendError, DisconnectedError can no longer be constructed by downstream code. That feels correct now, but can be fixed in non-API-breaking manner if needed.~ - no longer true after the last commit.
---

Is it worth it?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -94,20 +94,33 @@ impl fmt::Display for ActorError {
 
 impl std::error::Error for ActorError {}
 
-/// Reasons why sending a message to an actor can fail.
-#[derive(Debug)]
-pub enum SendError {
-    /// The channel's capacity is full.
-    Full,
-    /// The recipient of the message no longer exists.
-    Disconnected,
+/// Failures that can occur when sending a message to an actor.
+#[derive(Debug, Clone, Copy)]
+pub struct SendError {
+    recipient_name: &'static str,
+    reason: SendErrorReason,
+}
+
+impl SendError {
+    /// Get the name of the intended recipient.
+    pub fn recipient_name(&self) -> &'static str {
+        self.recipient_name
+    }
+
+    /// Get the reason why sending has failed.
+    pub fn reason(&self) -> SendErrorReason {
+        self.reason
+    }
 }
 
 impl fmt::Display for SendError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        match self {
-            SendError::Full => write!(f, "The channel's capacity is full."),
-            SendError::Disconnected => DisconnectedError.fmt(f),
+        let recipient_name = self.recipient_name;
+        match self.reason {
+            SendErrorReason::Full => {
+                write!(f, "The capacity of {}'s channel is full.", recipient_name)
+            },
+            SendErrorReason::Disconnected => DisconnectedError { recipient_name }.fmt(f),
         }
     }
 }
@@ -115,18 +128,29 @@ impl fmt::Display for SendError {
 impl std::error::Error for SendError {}
 
 /// The actor message channel is disconnected.
-#[derive(Debug)]
-pub struct DisconnectedError;
+#[derive(Debug, Clone, Copy)]
+pub struct DisconnectedError {
+    recipient_name: &'static str,
+}
 
 impl fmt::Display for DisconnectedError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "The recipient of the message no longer exists.")
+        write!(f, "The recipient of the message ({}) no longer exists.", self.recipient_name)
     }
 }
 
 impl std::error::Error for DisconnectedError {}
 
-impl<M> From<channel::TrySendError<M>> for SendError {
+/// Reasons why sending a message to an actor can fail.
+#[derive(Debug, Clone, Copy)]
+pub enum SendErrorReason {
+    /// The channel's capacity is full.
+    Full,
+    /// The recipient of the message no longer exists.
+    Disconnected,
+}
+
+impl<M> From<channel::TrySendError<M>> for SendErrorReason {
     fn from(orig: channel::TrySendError<M>) -> Self {
         match orig {
             channel::TrySendError::Full(_) => Self::Full,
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1001,4 +1034,27 @@ mod tests {
...
+        let error = full_actor.send(123).unwrap_err();
+        assert_eq!(error.to_string(), "The capacity of TestActor's channel is full.");
+        assert_eq!(
+            format!("{:?}", error),
+            r#"SendError { recipient_name: "TestActor", reason: Full }"#
+        );
+
+        system.shutdown().unwrap();
+
+        let error = stopped_actor.send(456usize).unwrap_err();
```
skywhale, line 1053: super-nit: type should be inferable?

```suggestion
        let error = stopped_actor.send(456).unwrap_err();
```

src/lib.rs
```diff
@@ -775,31 +808,31 @@ impl<M> Recipient<M> {
 
 pub trait SendResultExt {
     /// Don't return an `Err` when the recipient is at full capacity, run `func` in such a case instead.
-    fn on_full<F: FnOnce()>(self, func: F) -> Result<(), DisconnectedError>;
+    fn on_full<F: FnOnce(&'static str)>(self, func: F) -> Result<(), DisconnectedError>;
```
skywhale, line 811: Could you document what `&'static str` represents?

src/lib.rs
```diff
@@ -94,39 +94,63 @@ impl fmt::Display for ActorError {
...
+/// Failures that can occur when sending a message to an actor.
+#[derive(Debug, Clone, Copy)]
+pub struct SendError {
+    recipient_name: &'static str,
+    reason: SendErrorReason,
+}
+
+impl SendError {
+    /// Get the name of the intended recipient.
+    pub fn recipient_name(&self) -> &'static str {
```
skywhale, line 106: I remember having this conversation in some other PR, and we said we usually prefer making the fields `pub` instead of adding getters unless changing a value breaks an invariant?

skywhale (commented):
I agree that this is useful for debugging actor channel bloat issues.

Did you consider including the intended message instead, like crossbeam_channel does? Some pros and cons that I can think of:

Pros:
* We can inspect the content of the message that is being dropped in case of error
* Particularly useful for actors that take an enum message

Cons:
* Messages need to implement `Debug` + `Send`?
* There are cases multiple actors take the same message type e.g. `MediaFrame`.

src/lib.rs
strohel: The conversation was probably this one: https://github.com/tonarino/actor/pull/53#discussion_r634371138 (+ related #56).

I still don't have any clear opinion on this. Likely this case would fit to "compound, passive data structure" per [Rust API guidelines](https://rust-lang.github.io/api-guidelines/future-proofing.html#c-struct-private)?
> Public fields are most appropriate for struct types in the C spirit: compound, passive data structures. Otherwise, consider providing getter/setter methods and hiding fields instead.

This type being an _error_ type, I think I was subconsciously affected by https://matklad.github.io/2020/10/15/study-of-std-io-error.html and its praise of the `std::io::Error`.

Thinking about it more, I think I slightly lean towards making the fields public. The API stability much smaller concern for us than `std`.

src/lib.rs
strohel: Good catch, added.

strohel:
> Did you consider including the intended message instead, like crossbeam_channel does? Some pros and cons that I can think of:
>
> Pros:
> - We can inspect the content of the message that is being dropped in case of error
> - Particularly useful for actors that take an enum message
>
> Cons:
> - Messages need to implement Debug + Send?
> - There are cases multiple actors take the same message type e.g. MediaFrame.

That's a very valid remark. Agreed that seeing the message is even more useful than seeing the recipient.

The approach is made tricky by our ability to create `Recipient<N>` out of `Addr<M>` if `N` can be converted to `M`, introduced in #38.
1. `SendError` cannot simply contain `M` as `Recipient<N>` must have the `M` type erased.
2. If we _also_ require that `M` is convertible back to `N`, then `SendError` can contain `N`. But such a requirement feels weird and may be impossible to satisfy.
3. `SendError` could contain `Box<dyn Any>` (`Any` being `M`), but `Any` doesn't unfortunately allow "try downcasting to `Debug` trait". It only allows trying downcasting to a particular type, `M` in this case. That would make the whole `M` -> `N` type erasure a second class citizen.
4. `SendError` could contain `Box<dyn Debug>`, which would indeed require adding the `Debug` bound to `Actor::Message` (it is already `Send`) - and possibly also `Sync` (see below).

The ideal API would probably be: provide `Option<Box<dyn Debug>>` in `SendError` which would be `Some` if `M: Debug` and `None` otherwise. Maybe _specialization_ would allow this one day, but I see no way how to achieve that in current stable rust (it looks like one can "require" a bound, but cannot "detect optional bound being satisfied").

Anyways, option (4) looks viable, so I've submitted and alternate PR #60 but it suffers from its own problems.

It is possible that we've closed a lot of doors with #38 - it could be feasible to touch some of its core ideas. For example we may try to Box the messages and convert then only on the receiving side. That could allow option (1) above.

src/lib.rs
skywhale: Yeah, `io::Error` type comes with accessors, and I was also wondering whether it's where you sought for consistency. I don't feel strongly either way, but I wanted to ask your opinion, so I can start forming my habits.

src/lib.rs
skywhale: I see. Thank you for the clarification :)

skywhale:
Thank you for all the considerations. The PR is still worth merging as is, but I would also love to hear other Rust experts thoughts @bschwind @mcginty @efyang @slightknack

src/lib.rs
strohel: I ended up making the fields `pub`.

Rationale: if we ever want to contain original message `M` as a payload in the error, it would be more handy if it is pub so that users can move the message out of the error (without introducing special methods for it).

mcginty (approved):
This looks good! The only thing that felt sus is that the internal trait's `try_send` returns a `SendErrorReason` instead of a `SendError`, but I understand the logic behind it.

The layout is a lot like what we had to do in innernet for IO error contexts, so makes sense to me.

PR #60, opened 2021-07-27
For reference: Add Box<dyn fmt::Debug> to SendError, require Actor::Message: Debug

strohel:
An alternative to #59, doesn't build.

#### Add test for SendError

We're going to touch it, the tests will demonstrate the changes.

### Add Box<dyn fmt::Debug> to SendError, require Actor::Message: Debug

The problem: does not compile, because popular error libraries like `anyhow`
require the error to be `Sync`. Because we erase the `M` type from SendError,
we either require `M` to be `Sync` or we're not `Sync` at all.

(crosbeam::channel's `TrySendError<M>` doesn't suffer from such a problem,
it is `Sync` when M is `Sync` and vice-versa).

### DIFF
--- a/examples/echo.rs
+++ b/examples/echo.rs
@@ -44,9 +44,11 @@ type Chunk = Arc<[Sample; CHUNK_SAMPLES]>;
 const DELAY_CHUNKS: usize = 60;
 
 /// A chunk of samples that represents the "dry" (original, authentic) signal.
+#[derive(Debug)]
 struct DryChunk(Chunk);
 
 /// A chunk of sample that represents the "wet" (processed, edited) signal.
+#[derive(Debug)]
 struct WetChunk(Chunk);
 
 /// DryChunk converts to (unspecified) Chunk (but not the other way around).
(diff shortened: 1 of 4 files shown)


skywhale:
I haven't looked into deep enough, but could we do something along the line of `crossbeam` i.e. return `SendError<N: Debug + Sync>` for `M: Sync`, and `SendError<N: Debug>` for `M` (non-Sync)?

skywhale:
> I haven't looked into deep enough, but could we do something along the line of `crossbeam` i.e. return `SendError<N: Debug + Sync>` for `M: Sync`, and `SendError<N: Debug>` for `M` (non-Sync)? I guess not, without having two variants of `recipient()`?

Ah, you answered already in https://github.com/tonarino/actor/pull/59.

> The ideal API would probably be: provide Option<Box<dyn Debug>> in SendError which would be Some if M: Debug and None otherwise. Maybe specialization would allow this one day, but I see no way how to achieve that in current stable rust (it looks like one can "require" a bound, but cannot "detect optional bound being satisfied").

strohel:
> Ah, you answered already in #59.

Yep, we're basically hitting the same limitation. `SendError<N>` is parametrized by some `N` (which is convertible to `M`), but whether such `SendError` is `Sync` depends on `M`, not `N`, so we cannot (AFAICS) use the type system to "automatically propagate additional invariants from `M` to `SendError`."

strohel:
Closing, this served mainly as a demonstration companion to #59.

Issue #22, opened 2021-02-20
Support priority channels


skywhale:
Actors currently have a single ingress channel, and messages are all treated with the same priority. Sometimes we want to send a mix of important and non-important data to an actor, and we have seen that important data is dropped because non-important data is busy and occupies the channel.

We have `send_if_not_full()` to work around the issue, but it's a bit hacky. I want to introduce a more proper priority concept to the actor channels.

```
enum Priority {
  High,
  Low,
}

impl<M> Recipient<M> {
  fn send_with_priority(&self, message: N, priority: Priority) { ... }
}
```

PR #61, opened 2021-08-11
Port from crossbeam-channel to flume

strohel:
This is a prerequisite for message priorities #22 - crossbeam has built-in
randomization when receiving messages, while flume lets us control that
using a feature flag.

flume's API is largely the same as crossbeam's, only real difference is
that select() is constructed using builder pattern rather than a macro.

Actor own benchmark result change (crossbeam -> flume):
```
circular/circular (2 actors)
                        time:   [3.0926 ms 3.1223 ms 3.1528 ms]
                        thrpt:  [317.18 Kelem/s 320.28 Kelem/s 323.35 Kelem/s]
                 change:
                        time:   [-24.762% -21.495% -17.992%] (p = 0.00 < 0.05)
                        thrpt:  [+21.940% +27.380% +32.912%]
                        Performance has improved.
  1 (1.00%) high mild
  2 (2.00%) high severe
circular/circular (native CPU count - 1)
                        time:   [3.9945 ms 4.0224 ms 4.0493 ms]
                        thrpt:  [246.96 Kelem/s 248.61 Kelem/s 250.34 Kelem/s]
                 change:
                        time:   [-7.5891% -5.5178% -3.4030%] (p = 0.00 < 0.05)
                        thrpt:  [+3.5229% +5.8400% +8.2123%]
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  1 (1.00%) low severe
  6 (6.00%) low mild
  3 (3.00%) high mild
circular/circular (native CPU count)
                        time:   [20.821 ms 21.743 ms 22.666 ms]
                        thrpt:  [44.119 Kelem/s 45.993 Kelem/s 48.027 Kelem/s]
                 change:
                        time:   [-16.421% -10.076% -3.3897%] (p = 0.00 < 0.05)
                        thrpt:  [+3.5087% +11.205% +19.648%]
                        Performance has improved.
circular/circular (50 actors)
                        time:   [25.593 ms 27.095 ms 28.638 ms]
                        thrpt:  [34.918 Kelem/s 36.907 Kelem/s 39.073 Kelem/s]
                 change:
                        time:   [-31.525% -24.649% -17.014%] (p = 0.00 < 0.05)
                        thrpt:  [+20.502% +32.712% +46.038%]
                        Performance has improved.
```

That looks good! Measured on my laptop with frequency scaling and thermal
throttling, so there may be some noise.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -435,13 +435,13 @@ impl System {
                 .recv(&addr.control_rx, |msg| match msg {
                     Ok(control) => Received::Control(control),
                     Err(RecvError::Disconnected) => {
-                        unreachable!("We keep control_tx alive through addr, should no happen.")
+                        panic!("We keep control_tx alive through addr, should not happen.")
                     },
                 })
                 .recv(&addr.message_rx, |msg| match msg {
                     Ok(msg) => Received::Message(msg),
                     Err(RecvError::Disconnected) => {
-                        unreachable!("We keep message_tx alive through addr, should no happen.")
+                        panic!("We keep message_tx alive through addr, should not happen.")
                     },
                 });
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -42,7 +42,7 @@
 //! ```
 //!
 
-use crossbeam_channel::{self as channel, select, Receiver, Sender};
+use flume::{select::SelectError, Receiver, RecvError, Selector, Sender};
 use log::*;
 use parking_lot::{Mutex, RwLock};
 use std::{
@@ -67,6 +67,8 @@ pub enum ActorError {
     SystemStopped { actor_name: &'static str },
     /// The actor message channel is disconnected.
     ChannelDisconnected { actor_name: &'static str },
+    /// The actor control channel is disconnected.
+    ControlChannelDisconnected { actor_name: &'static str },
     /// Failed to spawn an actor thread.
     SpawnFailed { actor_name: &'static str },
     /// A panic occurred inside an actor thread.
@@ -81,6 +83,9 @@ impl fmt::Display for ActorError {
             },
             ActorError::ChannelDisconnected { actor_name } => {
                 write!(f, "The message channel is disconnected for the actor {}.", actor_name)
+            },
+            ActorError::ControlChannelDisconnected { actor_name } => {
+                write!(f, "The control channel is disconnected for the actor {}.", actor_name)
             },
             ActorError::SpawnFailed { actor_name } => {
                 write!(f, "Failed to spawn a thread for the actor {}.", actor_name)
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -82,6 +84,9 @@ impl fmt::Display for ActorError {
             ActorError::ChannelDisconnected { actor_name } => {
                 write!(f, "The message channel is disconnected for the actor {}.", actor_name)
             },
+            ActorError::ControlChannelDisconnected { actor_name } => {
```
skywhale, line 87: nit: `ControlChannel` is an internal concept, that the client of the library does not need to know about. You think it's okay to expose it as part of public API? or should we return `SystemStopped` since it's the only possible cause of `ControlChannelDisconnected`?

src/lib.rs
```diff
@@ -501,6 +530,15 @@ impl Deref for System {
     }
 }
 
+/// What can be received during one actor event loop.
+enum Received<M> {
```
skywhale, line 534: nit: Did you struggle with the naming? There is something a bit confusing about it. There is `Recipient` already, but `Received` is not received by `Recipient` but by `Actor`. Not that I can think of a better name, but `ActorMessage`, `ActorSelectLoopMessage`, something along that line?

skywhale (approved):
Looks great :)

The throughput even increased?!

src/lib.rs
strohel: > Did you struggle with the naming?

I did. :) I was able to move `Received<M>` into `run_actor_select_loop()` - it is its implementation detail. Do you think that is sufficient to resolve the confusing naming, or shall we still rename it?

src/lib.rs
skywhale: I think that makes sense :) `unreachable!()/expect()` would be more self-documenting.

src/lib.rs
skywhale: That's sufficient. Thank you @strohel :)

src/lib.rs
strohel: Pushed the change. It resulted in more code cleanups, so I'll wait for another round of review before merging.

src/lib.rs
```diff
@@ -454,15 +435,13 @@ impl System {
                 .recv(&addr.control_rx, |msg| match msg {
                     Ok(control) => Received::Control(control),
                     Err(RecvError::Disconnected) => {
-                        Received::Error(ActorError::ControlChannelDisconnected {
-                            actor_name: A::name(),
-                        })
+                        unreachable!("We keep control_tx alive through addr, should no happen.")
```
skywhale, line 438: typo: `should not happen`

skywhale: As per the suggestion from @mcginty on the other PR, let's switch to use `panic!()` here.

skywhale (approved):
LGTM++

src/lib.rs
strohel: :+1:

bschwind:
Sorry I didn't review this earlier. Those performance numbers look great, and the replacement implementation seems pretty straightforward to me. Post-merge LGTM!

strohel:
> Sorry I didn't review this earlier. Those performance numbers look great, and the replacement implementation seems pretty straightforward to me. Post-merge LGTM!

Thanks for post-merge review! :bow:

mcginty:
Same as Brian! Looks great, and yay for easy performance gainzzz.

PR #62, opened 2021-08-16
Add high priority message channel for each actor

### LINKED ISSUE #22: Support priority channels

strohel:
#### Remove Recipient::remaining_capacity()

Would have to accept extra parameter once we add priority channel.
That would be well possible, but perhaps this method in not really needed
once we have prio channels? Let's try without it first.

### Add high priority message channel for each actor

Actor trait has gained new optional method `priority(message: &M)` which
determines priority of each message before being sent.

`Addr` grows by one extra `flume::Receiver`. `Recipient` doesn't grow as its
`message_tx` payload is boxed.

#### Support specifying capacity independently for normal and high-priority channel

When porting portal to use priority channels, I've realized actor systems may
want to set capacity of the normal channel as low as 1 (for actors that cannot
keep up with the pace, but for which that doesn't matter).

Producers of their "bulk" low-priority messages will usually ignore their
channels being full. OTOH producers of their important high-priority messages
may like to fail when their priority channel is full (as ignoring that may
have dire consequences). It would be thus impractical if the priority channel
got similarly small capacity. (actor needs to finish at least one event loop
to get to process the high-prio messages).

#### Forward priority to inner actor in wrapping actors

In the `Timed` wrapper this is non-trivial, so some extra explanation docs
are added.

---
Fixes #22.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -811,7 +811,7 @@ impl<M> Clone for Recipient<M> {
 }
 
 impl<M> Recipient<M> {
-    /// Send a message to an actor. Returns [`SendError`] if the channel if full; does not block.
+    /// Send a message to an actor. Returns [`SendError`] if the channel is full; does not block.
     /// See [`SendResultExt`] trait for convenient handling of errors.
     pub fn send(&self, message: M) -> Result<(), SendError> {
         self.message_tx
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -229,6 +229,33 @@ impl<M> Context<M> {
     }
 }
 
+/// Capacity of actor's normal- and high-priority inboxes.
+/// For each inbox type, `None` signifies default capacity. Converts from [`usize`].
+#[derive(Clone, Copy, Debug, Default)]
+pub struct Capacity {
+    normal: Option<usize>,
+    high: Option<usize>,
+}
+
+impl Capacity {
+    /// Set capacity of the normal priority channel, and use default for the high priority one.
+    pub fn of_normal_priority(capacity: usize) -> Self {
+        Self { normal: Some(capacity), ..Default::default() }
+    }
+
+    /// Set capacity of the high priority channel, and use default for the normal priority one.
+    pub fn of_high_priority(capacity: usize) -> Self {
+        Self { high: Some(capacity), ..Default::default() }
+    }
+}
+
+/// Set capacity of both normal and high priority channels to the same amount of messages.
+impl From<usize> for Capacity {
+    fn from(capacity: usize) -> Self {
+        Self { normal: Some(capacity), high: Some(capacity) }
+    }
+}
+
 /// A builder for specifying how to spawn an [`Actor`].
 /// You can specify your own [`Addr`] for the Actor,
 /// the capacity of the Actor's inbox, and you can specify
@@ -237,7 +264,7 @@ impl<M> Context<M> {
 #[must_use = "You must call .spawn() or .block_on() to run this actor"]
 pub struct SpawnBuilder<'a, A: Actor, F: FnOnce() -> A> {
     system: &'a mut System,
-    capacity: Option<usize>,
+    capacity: Capacity,
     addr: Option<Addr<A>>,
     factory: F,
 }
@@ -245,9 +272,9 @@ pub struct SpawnBuilder<'a, A: Actor, F: FnOnce() -> A> {
 impl<'a, A: 'static + Actor<Context = Context<<A as Actor>::Message>>, F: FnOnce() -> A>
     SpawnBuilder<'a, A, F>
 {
-    /// Specify a capacity for the actor's receiving channel.
-    pub fn with_capacity(self, capacity: usize) -> Self {
-        Self { capacity: Some(capacity), ..self }
+    /// Specify a capacity for the actor's receiving channel. Accepts [`Capacity`] or [`usize`].
+    pub fn with_capacity(self, capacity: impl Into<Capacity>) -> Self {
+        Self { capacity: capacity.into(), ..self }
     }
 
     /// Specify an existing [`Addr`] to use with this Actor.
@@ -260,7 +287,7 @@ impl<'a, A: 'static + Actor<Context = Context<<A as Actor>::Message>>, F: FnOnce
     /// has stopped.
     pub fn run_and_block(self) -> Result<(), ActorError> {
         let factory = self.factory;
-        let capacity = self.capacity.unwrap_or(DEFAULT_CHANNEL_CAPACITY);
+        let capacity = self.capacity;
         let addr = self.addr.unwrap_or_else(|| Addr::with_capacity(capacity));
 
         self.system.block_on(factory(), addr)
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -1048,4 +1065,64 @@ mod tests {
...
+        system
+            .prepare(PriorityActor { received: Arc::clone(&received) })
+            .with_addr(addr)
+            .run_and_block()
+            .unwrap();
+
+        assert_eq!(
+            *received.lock(),
+            [10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+        );
```
strohel, line 1126: This nicely fails when one enables `eventual-fairness` on flume:
```
---- tests::message_priorities stdout ----
thread 'tests::message_priorities' panicked at 'assertion failed: `(left == right)`
  left: `[0, 10, 11, 12, 1, 2, 13, 14, 15, 16, 17, 3, 18, 4, 5, 19, 6, 7, 8, 9]`,
 right: `[10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9]`', src/lib.rs:1123:9
```

skywhale: Nice :) Thank you for confirming!

src/lib.rs
```diff
@@ -654,6 +660,12 @@ pub trait Actor {
     /// The name of the Actor - used only for logging/debugging.
     fn name() -> &'static str;
 
+    /// Determine priority of a `message` before it is sent to this actor.
```
skywhale, line 663: An alternative would be `Recipient::send_with_priority()`, which is more verbose. What were your thinkings to prefer this?

strohel: This is a good point, relates to discussion in https://github.com/tonarino/actor/issues/22#issuecomment-786974234.

The priority can be determined by 1) sender, 2) the message, 3) the receiver, or a combination of them. Which of these approaches should the API _encourage?_
1. sender-determined priority
   - pro: straightforward and flexible.
   - con: sender usually not in good position to determine priority, may not even know the destination actor (linked comment).
2. message-determined priority
   - pro: ability to reuse priorities among senders, receivers.
   - con: current Rust type system AFAIK does not allow "if the message implements WithPriority then use it, otherwise default to `Normal`". We either require all types implement some `WithPriority` or require callers to remember to use deticated send() method for messages that do (not) implement `WithPriority`.
   - con: downstream crates would not be able to implement `WithPriority` for types they do not own (orphan rule), would have to resort to newtypes.
   - con: what if 2 actors consume the same message type, but want to assign priorities differently? Actors would have to use newtypes to differentiate.
3. receiver-determined priority
   - pro: receiving actors should have the right context to determine message priorities.
   - pro: we're able to default to `Normal` priority without extra care and _at the same time_ allow actors to ensure certain messages receive `High` priority. I.e. it suffices to implement one function compared to checking all senders in (1).

This PR went for (c) **by default, but** it would be easy to add `Recipient::send_with_priority()` to it, which would override the receiver-assigned priority. That should even be a non-API-breaking change. The question is: should we add `send_with_priority()` right now in this PR, or leave it only as a future possiblity? Options:
1. Add it right in this PR.
2. Port our downstream crate portal to this PR before merging it and see if there's a need for `send_with_priority()`.
3. Merge this without `send_with_priority()`, leave it to future.

Perhaps option (2) would be the most wise one?

skywhale: Thank you for the thorough reasoning :) I now remember the conversation we had on the GitHub issue, and I agree that that receivers are in a better position to determine the priorities of the incoming messages. I also agree with your proposal of starting with this simple model, and only add overrides like `send_with_priority()` when they deem to be necessary.

skywhale (approved):
LGTM :)

bschwind (approved):
This is looking good, and I agree with the reasoning that the receiver decides the priority of a message. With this change, are we maintaining roughly the same performance we gained when switching to `flume` in the prerequisite PR?

strohel:
> With this change, are we maintaining roughly the same performance we gained when switching to flume in the prerequisite PR?

I looks like there is a slight regression compared to right after porting to flume:
```
circular/circular (2 actors)                                                                             
                        time:   [3.1650 ms 3.1949 ms 3.2268 ms]
                        thrpt:  [309.90 Kelem/s 312.99 Kelem/s 315.96 Kelem/s]
                 change:
                        time:   [+6.1280% +7.4680% +8.7130%] (p = 0.00 < 0.05)
                        thrpt:  [-8.0147% -6.9490% -5.7741%]
                        Performance has regressed.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
Benchmarking circular/circular (native CPU count - 1): Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 10.0s. You may wish to increase target time to 18.8s, enable flat sampling, or reduce sample count to 50.
circular/circular (native CPU count - 1)                                                                             
                        time:   [4.1088 ms 4.1343 ms 4.1591 ms]
                        thrpt:  [240.44 Kelem/s 241.88 Kelem/s 243.38 Kelem/s]
                 change:
                        time:   [+7.9763% +10.226% +12.621%] (p = 0.00 < 0.05)
                        thrpt:  [-11.207% -9.2773% -7.3871%]
                        Performance has regressed.
Found 5 outliers among 100 measurements (5.00%)
  3 (3.00%) low severe
  1 (1.00%) low mild
  1 (1.00%) high mild
circular/circular (native CPU count)                                                                            
                        time:   [22.316 ms 23.207 ms 24.119 ms]
                        thrpt:  [41.460 Kelem/s 43.090 Kelem/s 44.811 Kelem/s]
                 change:
                        time:   [+1.5387% +6.6061% +12.139%] (p = 0.02 < 0.05)
                        thrpt:  [-10.825% -6.1967% -1.5154%]
                        Performance has regressed.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
circular/circular (50 actors)                                                                            
                        time:   [19.662 ms 20.622 ms 21.649 ms]
                        thrpt:  [46.191 Kelem/s 48.492 Kelem/s 50.859 Kelem/s]
                 change:
                        time:   [-11.337% -4.4279% +2.8930%] (p = 0.25 > 0.05)
                        thrpt:  [-2.8117% +4.6330% +12.787%]
                        No change in performance detected.
Found 5 outliers among 100 measurements (5.00%)
  4 (4.00%) high mild
  1 (1.00%) high severe
```

I guess the reason is the select loop now has to check the priority channel first, i.e. one extra operation per message received. The regression of adding the channel is fortunately lower than what we gained by using `flume` in the first place. I have to say that the benches are rather fluctuating at my laptop (in one run I got `51.5 Kelem/s` ~ +17% for `circular/circular (native CPU count)`), so we should take them with a grain of salt.

---
I have pushed one extra commit, to also update wrapping actors.[1] With that it should be hopefully complete.

[1] This illustrates one reason I don't usually like trait methods with default implementations: it is too easy to forget implementing them, especially if they are added later. OTOH I didn't like _not_ adding the default implementation either: having the default allows actor system not care at all about the priority if they don't want to.

PR #63, opened 2021-08-19
Support specifying capacity independently for normal and high-priorit…

strohel:
…y channel

When porting portal to use priority channels, I've realized actor systems may
want to set capacity of the normal channel as low as 1 (for actors that cannot
keep up with the pace, but for which that doesn't matter).

Producers of their "bulk" low-priority messages will usually ignore their
channels being full. OTOH producers of their important high-priority messages
may like to fail when their priority channel is full (as ignoring that may
have dire consequences). It would be thus impractical if the priority channel
got similarly small capacity. (actor needs to finish at least one event loop
to get to process the high-prio messages).

---
There is quite some freedom in how to design API of this, so I thought it would be handy to run this as a separate PR that targets #62.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -229,6 +229,33 @@ impl<M> Context<M> {
     }
 }
 
+/// Capacity of actor's normal- and high-priority inboxes.
+/// For each inbox type, `None` signifies default capacity. Converts from [`usize`].
+#[derive(Clone, Copy, Debug, Default)]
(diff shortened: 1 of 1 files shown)


skywhale (approved):
LGTM

This makes a lot of sense :)

strohel:
Thanks for review, merging this to the parent PR (i.e. not yet to `main`).

PR #64, opened 2022-02-15
Add `priority` field to `SendError`, `DisconnectedError`

strohel:
#### Move `name` field from `Recipient` to internal `MessageSender`

This is a tiny struct size optimization, because multiple Recipients share a single MessageSender
through an Arc.

But more importantly this allows us to implement the next commit (add priority to error types) more
cleanly.

#### Add `priority` field to `SendError`, `DisconnectedError`

This should help with diagnosing actor channel full errors.

Also changes signature of the callback accepted by SendErrorExt::on_full() method, which is
a breaking change.

#### Fix non-determinism in errors() test

Thanks @bschwind for pointing-out zero-capacity flume channels are special.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -94,6 +94,8 @@ impl std::error::Error for ActorError {}
 pub struct SendError {
     /// The name of the intended recipient.
     pub recipient_name: &'static str,
+    /// The priority assigned to the message that could not be sent.
+    pub priority: Priority,
     /// The reason why sending has failed.
     pub reason: SendErrorReason,
 }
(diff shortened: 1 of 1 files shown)


skywhale (approved):
LGTM :)

Would you increment semvar in a separate PR?

src/lib.rs
```diff
@@ -1087,10 +1100,13 @@ mod tests {
         let stopped_actor = system.spawn(TestActor).unwrap().recipient();
 
         let error = full_actor.send(123).unwrap_err();
```
bschwind, line 1102: It seems CI is failing on this line. I'm curious why it passed before. We're creating a zero-capacity channel, and flume's documentation has this to say:

> Like std::sync::mpsc, flume supports ‘rendezvous’ channels. A bounded queue with a maximum capacity of zero will block senders until a receiver is available to take the value. You can imagine a rendezvous channel as a [‘Glienicke Bridge’](https://en.wikipedia.org/wiki/Glienicke_Bridge)-style location at which senders and receivers perform a handshake and transfer ownership of a value.

https://docs.rs/flume/0.10.11/flume/fn.bounded.html

strohel:
> Would you increment semvar in a separate PR?

I plan to do 2 more small PRs and then cut a new release (semver-incompatible, i.e. 0.8.0) - it seems that cargo-release will bump the version for us [judging from the last time](https://github.com/tonarino/actor/commit/60022c4074).

src/lib.rs
strohel: This that was non-deterministic, failing in like 20% of the cases locally.

I didn't know why... until you've looked up that capacity zero is special, which I was not aware of! :) Thank you, I'll change the test to use non-zero capacity.

strohel: Resolved in e814850.

PR #65, opened 2022-02-16
Move `.recipient()` from `Addr` to `Recipient`, Capacity fields public

strohel:
#### Really make the errors() test deterministic

I claimed this in the previous PR, but it wasn't true: if the actor spawned quick enough the test
would fail. What _was_ true is that I've ran the testsuite hundreds of times in a loop, and it
didn't fail. Some strange coincidence?

Make it truly deterministic by not spawning the actor at all (test using a naked `Addr`).

#### Move `.recipient()` from `Addr` to `Recipient`

This makes the API more flexible as we now can call `.recipient()` multiple times in a row on the
same value, allowing one more M -> N message conversion in each step.

Previously, we thought it is too wild to create "arbitrarily long conversion chain", so we've put
`.recipient()` on `Addr`, only allowing one such conversion.

But with the advent of actor wrappers like `Timed<A>`, this first conversion is "consumed" by the
wrapper itself, limiting their users.

The method name `Recipient::recipient()` seems a bit weird, but there's a high value in keeping the
current usage working without any changes (because Addr derefs to Recipient), and no obviously
better name came into my mind.

#### Make `Capacity` fields (high, normal) public

This is an API omission discovered when we tried to set both normal and priority inbox sizes to
different and non-default values.

I argue that `Capacity { normal: Some(3), high: Some(9) }` is more explicit than
`Capacity::new(3, 9)`, hence making the fields pub instead of adding a constructor.

---

Sorry for cramming this into one PR (the commits are still atomic), [I miss Gerrit](https://en.wikipedia.org/wiki/Carthago_delenda_est)! ;)

### DIFF
--- a/examples/delay_actor.rs
+++ b/examples/delay_actor.rs
@@ -33,8 +33,10 @@ fn main() -> Result<(), Error> {
 
     let mut system = System::new("Example Timer System");
 
-    let consumer =
-        system.spawn(Timed::new(FinalConsumer { started_at: Instant::now() }))?.recipient();
+    let consumer = system
+        .prepare(Timed::new(FinalConsumer { started_at: Instant::now() }))
+        .with_capacity(6)
+        .spawn()?;
 
     let now = Instant::now();
     consumer.send_recurring(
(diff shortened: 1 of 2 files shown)


skywhale (approved):
LGTM

I also couldn't come up with a better name than `recipient` here.

bschwind (approved):
LGTM!

Only thing I might request is showing off new code that this `recipient` change allows, which wasn't possible before the change.

strohel:
> Only thing I might request is showing off new code that this `recipient` change allows, which wasn't possible before the change.

That's a good point. That will also server as a form of a compile-time unit-test of the API. I've added it in
`Add example of .recipient() chaining to delay_actor example` a772f09e5

With that, I think I'll merge this after my lunch and cut a new release.

PR #66, opened 2022-02-18
Allow specifying default capacities on actors themselves

### LINKED ISSUE #70: Let `Actor`s specify their default `Addr` capacities
Currently, if a user wants to create an actor with non-default (normal and/or high-priority) inbox capacities, they need to either create their `addr` first using `Addr::with_capacity()` or spawn them using `system.prepare[_fn]().with_capacity()`.

However, the actors themselves often have enough information to determine their best (or at least good default) inbox capacities, so let's allow them doing so! This has a potential to noticeably simplify "actor system setup" codebase sections.

`Addr` already has actor generic type parameter, so it can access static methods and constants defined on it in its constructor.

Implementation-wise this would be about adding either an associated constant to `Actor` trait to determine default capacity (if Rust allows having defaults for associated constants), or a static method (without `&self`) to the `Actor` trait with default implementation. In both cases we likely don't want to add any more _mandatory_ items that Actors need to implement. Change `impl Default for Addr` to use these method(s)/constant(s).

Another thing left to bikeshed is whether this should be a single constant/method returning `Capacity`, or 2 separate methods/constants for normal, high-prio capacities returning `usize`. Notice the `DEFAULT_CHANNEL_CAPACITY` constant is not currently public.

CC @bschwind @skywhale @PabloMansanet.

strohel:
I'm myself not convinced about this, but publishing for feedback anyway.

Resolves #70 (I apparently forgot about this PR when writing the issue; it even shows that defaults for associated constants are possible/stable).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -75,8 +75,8 @@ pub mod timed;
 #[cfg(test)]
 pub mod testing;
 
-// Default capacity for channels unless overridden by `.with_capacity()`.
-static DEFAULT_CHANNEL_CAPACITY: usize = 5;
+/// Capacity of the control channel (used to deliver [Control] messages).
+const CONTROL_CHANNEL_CAPACITY: usize = 5;
 
 #[derive(Debug)]
 pub enum ActorError {
@@ -298,7 +298,7 @@ impl<M> Context<M> {
 }
 
 /// Capacity of actor's normal- and high-priority inboxes.
-/// For each inbox type, `None` signifies default capacity. Converts from [`usize`].
+/// For each inbox type, `None` signifies default capacity of given actor. Converts from [`usize`].
 #[derive(Clone, Copy, Debug, Default)]
 pub struct Capacity {
     pub normal: Option<usize>,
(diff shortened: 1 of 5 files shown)


bschwind:
@strohel I've rebased this branch on the latest `main`

bschwind:
Related to #70

strohel:
Thanks, @bschwind - yeah we should get this done. Looks like you're on it?

strohel:
@bschwind I think we can just merge this and cut a new version, WDYT?

src/lib.rs
```diff
@@ -341,6 +341,8 @@ impl<'a, A: 'static + Actor<Context = Context<<A as Actor>::Message>>, F: FnOnce
     SpawnBuilder<'a, A, F>
 {
     /// Specify a capacity for the actor's receiving channel. Accepts [`Capacity`] or [`usize`].
+    ///
+    /// Ignored when `.with_addr()` is used at the same time.
```
skywhale, line 345: Thank you for adding this documentation :)

skywhale (approved):
LGTM

The changes make a lot of sense to me :)

bschwind (approved):
LGTM!

I don't believe this is a breaking change, so we can just do a patch release, right?

strohel:
> I don't believe this is a breaking change, so we can just do a patch release, right?

I think/hope so, so let me jut cut it.

PR #69, opened 2022-07-06
Fix clippy 1.62 lint

strohel:


PabloMansanet:
LGTM!

This reminds me of a story of a time long past, where I got in trouble for refactoring a C++ function that looked like this:

```c++
void foo() {
   auto* p = new BigClass();
   auto* q = new AnotherBigClass();
}
```

into this:
```c++
void foo() {
    new BigClass();
    new AnotherBigClass();
}
```

If you're going to leak, why not make it obvious right? :P

PR #73, opened 2023-02-24
Fix timed messages for busy actors

### LINKED ISSUE #72: Delayed/recurring messages are not delivered to busy actors
When a `Timed<>` actor does something like
```rust
fn handle(&mut self, context, message) {
    sleep(some time);
    context.myself.send_now(some message);
}
```

Then delayed or recurring messages are never delivered to it (no matter what priority they have), because `Timed` wrapper only processes them in `deadline_exceeded()`, which is never called (because the queue in never empty in actors event loop handling).

strohel:
### Add test showcasing delayed messages not delivered to busy actors 

This test fails (as of this commit) with
```
---- timed::tests::recurring_messages_for_busy_actors stdout ----
thread 'timed::tests::recurring_messages_for_busy_actors' panicked at 'assertion failed: `(left == right)`
  left: `[1, 3, 3]`,
 right: `[1, 2, 3, 2, 3]`', src/timed.rs:318:9
```

I.e. messages of type 2 (delayed) are never delivered.

### Process delayed/recurring messages also for busy actors

This fixes the recurring_messages_for_busy_actors test, and should fix #72.

---
_Please give the contained logic thorough review scrutiny, this is a central thing._

### DIFF
--- a/src/timed.rs
+++ b/src/timed.rs
@@ -249,3 +249,73 @@ enum Payload<M> {
     Delayed { message: M },
     Recurring { factory: Box<dyn FnMut() -> M + Send>, interval: Duration },
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::System;
+    use std::{
+        sync::{Arc, Mutex},
+        thread,
+    };
+
+    struct TimedTestActor {
+        received: Arc<Mutex<Vec<usize>>>,
+    }
+
+    impl Actor for TimedTestActor {
+        type Context = TimedContext<Self::Message>;
+        type Error = ();
+        type Message = usize;
+
+        fn name() -> &'static str {
+            "TimedTestActor"
+        }
+
+        fn handle(&mut self, context: &mut Self::Context, message: usize) -> Result<(), ()> {
+            {
+                let mut guard = self.received.lock().unwrap();
+                guard.push(message);
+            }
+
+            // Messages 1 or 3 are endless self-sending ones, keep to loop spinning.
+            if message == 1 || message == 3 {
+                thread::sleep(Duration::from_millis(100));
+                context.myself.send_now(3).unwrap();
+            }
+
+            Ok(())
+        }
+
+        fn started(&mut self, context: &mut Self::Context) {
+            context
+                .myself
+                .send_recurring(
+                    || 2,
+                    Instant::now() + Duration::from_millis(50),
+                    Duration::from_millis(100),
+                )
+                .unwrap()
+        }
+    }
+
+    #[test]
+    fn recurring_messages_for_busy_actors() {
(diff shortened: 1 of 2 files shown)


src/timed.rs
```diff
@@ -249,3 +260,73 @@ enum Payload<M> {
...
+            "TimedTestActor"
+        }
+
+        fn handle(&mut self, context: &mut Self::Context, message: usize) -> Result<(), ()> {
+            {
+                let mut guard = self.received.lock().unwrap();
+                guard.push(message);
+            }
+
+            // Messages 1 or 3 are endless self-sending ones, keep to loop spinning.
```
skywhale, line 292: typo?

skywhale (approved):
Thank you for finding the bug and addressing quickly. Code LGTM

strohel:
I've tested this in https://github.com/tonarino/portal/pull/2293 and it's looking good!

@skywhale (or anybody else) I've changed order of processing slightly, added/extended docs and fixed the typo, PTAL.

Once we merge this I'll do a release.

src/timed.rs
```diff
@@ -136,6 +136,12 @@ impl<M: Send + 'static, A: Actor<Context = TimedContext<M>, Message = M>> Timed<
             };
 
             // Let inner actor do its job.
+            //
+            // Alternatively, we could send an `Instant` message to ourselves.
```
skywhale, line 140: Thank you for additional documentation and agree with the current strategy.

src/timed.rs
```diff
@@ -161,24 +167,24 @@ impl<M: Send + 'static, A: Actor<Context = TimedContext<M>, Message = M>> Actor
...
             TimedMessage::Recurring { factory, fire_at, interval } => {
-                QueueItem { fire_at, payload: Payload::Recurring { factory, interval } }
+                self.queue
+                    .push(QueueItem { fire_at, payload: Payload::Recurring { factory, interval } });
             },
         };
 
-        self.queue.push(item);
+        // Process any expired items in the queue. In case that the actor is non-stop busy (there's
+        // always a message in its queue, perhaps because it sends a message itself in handle()),
```
skywhale, line 184: a message "to" itself

skywhale (approved):
LGTM++

bschwind (approved):
LGTM!

As an aside, did you notice any change in the speed of the benchmarks?

src/timed.rs
strohel:
```suggestion
        // always a message in its queue, perhaps because it sends a message to itself in handle()),
```

strohel:
> As an aside, did you notice any change in the speed of the benchmarks?

We unfortunately don't have any benches that used the `Timed` wrapper (and this change is strictly contained to he timed wrapper).

PR #78, opened 2023-10-18
Fix clippy 1.73 lints, update lint configuration

strohel:
#### timed::QueueItem: Implement partial_cmp() using cmp(), not the other way around

Fixes clippy 1.73 lint.

This different order makes sense, as it avoids an expect().

#### Clippy lints update: `all` is implied, add `todo`

I must have been mistaken when originally introducing `clippy:all`, because these are, by definition, lints that are enabled without any configuration: https://doc.rust-lang.org/stable/clippy/usage.html#lint-configuration

Add a lint that catches `todo!()` (but allows `panic!()`, `unreachable!()`, `unimplemented!()`) - you're encouraged to use it during development as a reminder for stuff to finish before a change can be merged.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,4 +1,4 @@
-#![warn(clippy::all, clippy::clone_on_ref_ptr)]
+#![warn(clippy::clone_on_ref_ptr, clippy::todo)]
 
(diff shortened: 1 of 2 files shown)


bschwind (approved):
LGTM!

PR #80, opened 2024-01-12
Timed infinite loop fix

### LINKED ISSUE #79: Inifinite loop when `Timed` actor's `handle()` is slower than recurring message interval
Caused by this code
```rust
    /// Process any pending messages in the internal queue, calling wrapped actor's `handle()`.
    fn process_queue(&mut self, context: &mut <Self as Actor>::Context) -> Result<(), A::Error> {
        // Handle all messages that should have been handled by now.
        let now = Instant::now();
        while self.queue.peek().map(|m| m.fire_at <= now).unwrap_or(false) {
            let item = self.queue.pop().expect("heap is non-empty, we have just peeked");

            let message = match item.payload {
                Payload::Delayed { message } => message,
                Payload::Recurring { mut factory, interval } => {
                    let message = factory();
                    self.queue.push(QueueItem {
                        fire_at: item.fire_at + interval,
                        payload: Payload::Recurring { factory, interval },
                    });
                    message
                },
            };

            // Let inner actor do its job.
            //
            // Alternatively, we could send an `Instant` message to ourselves.
            // - The advantage would be that it would go into the queue with proper priority. But it
            //   is unclear what should be handled first: normal-priority message that should have
            //   been processed a while ago, or a high-priority message that was delivered now.
            // - Disadvantage is we could easily overflow the queue if many messages fire at once.
            self.inner.handle(&mut TimedContext::from_context(context), message)?;
        }

        Ok(())
    }
```
Combined with the fact that we call this from `handle()`.

I think that we should:
- send message to ourselves instead of calling `inner.handle()`, so that it goes through the actor loop and gets proper priority
- only process one message at a time: change `while let` to `if let`
  - this would resolve the concern in the TODO in the code that we 

CC @goodhoko.

strohel:
### Add test for issue #79

Not currently failing, just demonstrates the unfortunate behavior.

## timed: fix #79 by enqueueing (rather than handling) delayed messages when due

Also rename `fire_at` to `enqueue_at` to be explicit about the fact.

This is a trade-off that prevents 2 sorts of bad behavior:
- actors with send-sending messages never handling any delayed/recurring messages (#72)
- actors with recurring messages that are slower to handle than their interval eventually not processing any outside messages (#79)

See the tweaked tests for the change of behavior.

---
Checking how the test results change commit-by-commit should give a good story.

### DIFF
--- a/src/timed.rs
+++ b/src/timed.rs
@@ -3,6 +3,7 @@
 //! To apply this to a given (receiving) actor:
 //! * Use [`TimedContext<Self::Message>`] as [`Actor::Context`] associated type.
 //!   * Such actors cannot be spawned unless wrapped, making it impossible to forget wrapping it.
+//! * Wrapped actor's `Error` must implement [`From<SendError>`].
 //! * Wrap the actor in [`Timed`] before spawning.
 //!
 //! The wrapped actor will accept [`TimedMessage<M>`] with convenience conversion from `M`.
@@ -10,9 +11,8 @@
 //! `send_delayed()`, `send_recurring()`.
 //!
 //! Once accepted by the actor, delayed and recurring messages do not occupy place in actor's
-//! channel inbox, they are placed to internal queue instead. Due to the design, delayed and
-//! recurring messages have always lower priority than instant messages when the actor is
-//! saturated.
+//! channel inbox, they are placed to internal queue instead. When delayed/recurring message become
+//! due, they go through the actor's regular inboxes (subject to prioritization).
 //!
 //! See `delay_actor.rs` example for usage.
 
@@ -24,11 +24,11 @@ use std::{
     time::{Duration, Instant},
 };
 
-/// A message that can be delivered now, at certain time and optionally repeatedly.
+/// A message that can be enqueued now, at certain time and optionally repeatedly.
 pub enum TimedMessage<M> {
     Instant { message: M },
-    Delayed { message: M, fire_at: Instant },
-    Recurring { factory: Box<dyn FnMut() -> M + Send>, fire_at: Instant, interval: Duration },
+    Delayed { message: M, enqueue_at: Instant },
+    Recurring { factory: Box<dyn FnMut() -> M + Send>, enqueue_at: Instant, interval: Duration },
 }
 
 /// This implementation allows sending direct unwrapped messages to wrapped actors.
@@ -43,19 +43,19 @@ pub trait RecipientExt<M> {
     /// Send a `message` now. Convenience to wrap message in [`TimedMessage::Instant`].
     fn send_now(&self, message: M) -> Result<(), SendError>;
 
-    /// Send a `message` to be delivered later at a certain instant.
-    fn send_timed(&self, message: M, fire_at: Instant) -> Result<(), SendError>;
+    /// Send a `message` to be enqueued later at a certain instant.
+    fn send_timed(&self, message: M, enqueue_at: Instant) -> Result<(), SendError>;
 
-    /// Send a `message` to be delivered later after some time from now.
+    /// Send a `message` to be enqueued later after some time from now.
     fn send_delayed(&self, message: M, delay: Duration) -> Result<(), SendError> {
         self.send_timed(message, Instant::now() + delay)
     }
 
-    /// Schedule sending of message at `fire_at` plus at regular `interval`s from that point on.
+    /// Schedule sending of message at `enqueue_at` plus at regular `interval`s from that point on.
     fn send_recurring(
         &self,
         factory: impl FnMut() -> M + Send + 'static,
-        fire_at: Instant,
+        enqueue_at: Instant,
         interval: Duration,
     ) -> Result<(), SendError>;
 }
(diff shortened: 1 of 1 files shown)


src/timed.rs
```diff
@@ -301,38 +303,78 @@ mod tests {
                 context.myself.send_now(3).unwrap();
             }
 
-            Ok(())
-        }
+            // Message 2 is a recurring one, sleep based on a paremeter.
```
bschwind, line 306:
```suggestion
            // Message 2 is a recurring one, sleep based on a parameter.
```

src/timed.rs
```diff
@@ -195,14 +199,11 @@ impl<M: Send + 'static, A: Actor<Context = TimedContext<M>, Message = M>> Actor
...
-                A::priority(message)
-            },
-            // Recurring message is only received once, the recurring instances go through the
-            // internal queue (and not actor's channel). Assign high priority to the request to
-            // set-up the recurrent sending.
-            TimedMessage::Recurring { .. } => Priority::High,
+            // Use underlying message priority for instant messages.
+            TimedMessage::Instant { message } => A::priority(message),
+            // Recurring and Delayed messages are only added to the queue when handled, and then go
+            // through actors priority inboxes again when actually enqueued.
```
bschwind, line 205: I'm having trouble visualizing the message flow based on this comment.

> only added to the queue when handled

In my mind, "handled" means the actor has run the `handle()` function on the message, so it's already out of the queue.

Is it possible to be more specific in the language here about the flow of the message from when `send_delayed` is called to when it actually ends up in the actor `handle()` function?

bschwind (approved):
LGTM, though I have a small question, and a spelling fix.

src/timed.rs
```diff
@@ -111,50 +111,52 @@ pub struct Timed<A: Actor> {
...
 
     /// Process any pending messages in the internal queue, calling wrapped actor's `handle()`.
-    fn process_queue(&mut self, context: &mut <Self as Actor>::Context) -> Result<(), A::Error> {
-        // Handle all messages that should have been handled by now.
-        let now = Instant::now();
-        while self.queue.peek().map(|m| m.fire_at <= now).unwrap_or(false) {
+    fn process_queue(&mut self, context: &mut <Self as Actor>::Context) -> Result<(), SendError> {
+        // If the message on top of the queue is due, send it to the regular actor queue.
+        // No problem if there are multiple such messages, it's handle() will call process_queue()
+        // again.
```
goodhoko, line 126: It's not clear to me how 

> it's handle() will call process_queue() again.

As far as I see, the inner actor doesn't have any knowledge of being wrapped in `Timed` and thus can't call outer's methods. 

EDIT: I now see, how this is meant. We're sending the due message to the outer as `Instant` effectively calling outer's `handle` which indeed then calls `process_queue()`.

Make me think if this wording would be clearer:

```suggestion
        // If the message on top of the queue is due, send it to ourselves as `Instant` to enqueue it in the regular actor queue.
        // No problem if there are multiple such messages, next handle() will call process_queue()
        // again.
```

src/timed.rs
goodhoko: Agree. What about

```suggestion
            // These priorities apply to the set-up of Delayed and Recurring messages and when want to handle that pronto.
            // The resulting inner message then comes back as `Instant` and is prioritized per its internal priority.
```

goodhoko (approved):
Thanks for materialising our long discussion into an actual fix @strohel!

Behaviorally this looks all correct! I only have some suggestion for improving the comments. Let's give them good amount of attention as they're important for understanding the convoluted logic of `Timed` wrapper. .)

src/timed.rs
strohel: @goodhoko you suggestion reads very good to me using it (with very small tweaks)

src/timed.rs
strohel: @goodhoko thanks for the suggestion, also using (with tiny wordsmith)

strohel:
@bschwind @goodhoko I've pushed a fixup commit that extends comments as suggested (+ also extends the docs directly on `TimedMessage` variants).

goodhoko (approved):
This is perfect! Especially the doc comments on TimedMessage variants are very helpfull. Good to merge IMO!

I suppose this can be released with a mere patch bump. Or do we consider the change in delivery order to be breaking?

strohel:
> I suppose this can be released with a mere patch bump. Or do we consider the change in delivery order to be breaking?

Good point. The rename of the `TimedMessage::Delayed` `fire_at` field to `enqueue_at` is technically semver-incompatible, so we should bump the minor version. It will at make it easier for donwstreams to downgrate `actor` version shuold they need it.

I'll prepare portal version that uses this PR before merging to get some real-life testing. It doesn't technically require any code changes (the renamed fields are not directly used). Also opportunity for @bschwind to check clarity of comments of the latest version.

goodhoko:
> The rename of the TimedMessage::Delayed fire_at field to enqueue_at is technically semver-incompatible


Ahaaa! I assumed that the enum field is private by default (as are struct fields) but actually enum variants and their fields always share the visibility of the enum itself. I.e. the field is part of the crate's public API.

> so we should bump the minor version. It will at make it easier for donwstreams to downgrate actor version shuold they need it.

Per the [semver spec](https://semver.org/) if it's backward incompatible we should bump the *major* version. Minor version bumps are for additions (backward compatible changes). It feels excessive but hey, it's just a version number.

goodhoko:
@strohel Got it. I didn't realize actor is still pre-1.0. Thanks!

strohel:
Tonari internal note: if we decide to revive this, the logic in https://github.com/tonarino/portal/pull/3692 may stop working.

PR #82, opened 2024-05-16
Missing address warning

mbernat:
This PR adds a warning when a `SpawnBuilder` is used without an address and with a default capacity. Such usage is often erroneous and it's hard to debug why such an actor is not receiving messages coming to an unassigned address.

Alternatives:
* It might be preferable to make this situation a hard error (ideally at compile-time) but I wasn't sure whether we want that.
* Another possibility is to add a warning on the sending side, if an address wasn't connected to an actor.

The other changes in this PR are fixing clippy.

### DIFF
--- a/benches/pub_sub.rs
+++ b/benches/pub_sub.rs
@@ -4,7 +4,7 @@ use std::time::{Duration, Instant};
 use tonari_actor::{Actor, Context, Event, Recipient, System};
 
 #[derive(Debug, Clone)]
-struct StringEvent(String);
+struct StringEvent(());
 
 impl Event for StringEvent {}
 
@@ -45,7 +45,7 @@ impl Actor for PublisherActor {
             PublisherMessage::PublishEvents => {
                 let start = Instant::now();
                 for _i in 0..self.iterations {
-                    context.system_handle.publish(StringEvent("hello".to_string()))?;
+                    context.system_handle.publish(StringEvent(()))?;
                 }
                 let elapsed = start.elapsed();
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1037,7 +1037,6 @@ impl<M: Into<N>, N> SenderTrait<M> for Arc<dyn SenderTrait<N>> {
 #[cfg(test)]
 mod tests {
     use std::{
-        rc::Rc,
         sync::atomic::{AtomicU32, Ordering},
         time::Duration,
     };
@@ -1103,7 +1102,7 @@ mod tests {
     #[test]
     fn send_constraints() {
         #[derive(Default)]
-        struct LocalActor(Rc<()>);
+        struct LocalActor(());
         impl Actor for LocalActor {
             type Context = Context<Self::Message>;
             type Error = ();
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -353,7 +353,11 @@ impl<'a, A: 'static + Actor<Context = Context<<A as Actor>::Message>>, F: FnOnce
     pub fn run_and_block(self) -> Result<(), ActorError> {
         let factory = self.factory;
         let capacity = self.capacity;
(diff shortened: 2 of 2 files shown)


PabloMansanet:
Hmmm... I don't know if this is the right solution.

I think we're projecting too much of the pattern we use in the `tonari` codebase, and expecting it to be the "correct" way.

The way the Actor API is designed, it's completely OK to spawn an actor with no predefined address. I'd say the API returning `Addr` heavily implies you're not expected to pass an address in, except when you have a reason to do so.

Warning for a valid usage of the API doesn't seem correct, even if it makes sense in the context of the use we've been giving the actor framework ourselves.

Instead, the solution I was thinking about was similar to your second "alternative": it would be to hard-error when a message is *sent* to an address that's not associated to any actor, either because it wasn't passed in to the SpawnBuilder, or because it didn't come from a default `spawn`. Haven't looked into it enough to come up with a way to implement this.

strohel (commented):
I'm also somewhat unsure about this change, a runtime warning would be easily missed for something that is likely a programmer oversight.

I'd prefer if we explore API-based prevention of this instead. Perhaps the there can be "staged" or split builder structs, and requiring specification of either (exclusively):
- address
- capacity
- "make and address with default capacity for me"

That could fix another caveat of the current API: if you specify both address _and_ capacity, the capacity is silently ignored and the one from address is used instead.

strohel:
> Another possibility is to add a warning on the sending side, if an address wasn't connected to an actor.

> Instead, the solution I was thinking about was similar to your second "alternative": it would be to hard-error when a message is _sent_ to an address that's not associated to any actor, either because it wasn't passed in to the SpawnBuilder, or because it didn't come from a default `spawn`.

This may be hard to do, because it _is_ a valid use-case to create an Addr, send to messages to it, and only then spawn an actor that processes these messages.

Otherwise it would be tricky to create some actor systems with circular dependencies.

PabloMansanet:
> > Another possibility is to add a warning on the sending side, if an address wasn't connected to an actor.
> 
> > Instead, the solution I was thinking about was similar to your second "alternative": it would be to hard-error when a message is _sent_ to an address that's not associated to any actor, either because it wasn't passed in to the SpawnBuilder, or because it didn't come from a default `spawn`.
> 
> This may be hard to do, because it _is_ a valid use-case to create an Addr, send to messages to it, and only then spawn an actor that processes these messages.
> 
> Otherwise it would be tricky to create some actor systems with circular dependencies.

That's a good point.

I wonder if we should zero in on the actual issue that we had: the queue *overflowed* because there was no actor handling it. Maybe it suffices to add a special warning/error message to the queue overflow, in the cases that Addr has never been associated to a spawned actor. 

Something like "Queue overflowed. No actor has been spawned with than address, make sure you didn't mix up the addresses", or similar.

mbernat:
Yeah, I tried doing something on the address side and it's non-trivial. Also, our current design is in a kind of a local optimum where `Addr` is not technically an address but contains both ends of the channel but it publicly derefs to the sender sides (`Recipient`) and privately exposes the receiver sides only to the actor framework. But you could probably still misuse it by assigning the same address to two different actors.

It's hard to see how to improve on this design without major changes. We'd need to attach some extra state to it using interior mutability and it feels much worse than what we have right now.

strohel:
Right, I think the only change that could be worth it _now_ would be on the `SpawnBuilder` side.

More specifically, we would have a 2-stage builder. The first stage without any `addr` or `capacity`, and the second stage with `addr`. The only way to go from the first stage to the second would be to use (exactly once)
- `with_addr(addr)`
- `with_capacity(capacity)`
- `with_default_capacity()`

This would be indeed semver-breaking, but most (if not all) valid usages should still compile without any changes.

mbernat:
Closing this in favor of new PR with the API proposed by @strohel .

PR #83, opened 2024-05-20
Make spawn builder harder to misuse

mbernat:
This is a follow-up to #82 to make it a hard error when the user forgets to provide an address to an actor.

This breaks the API because now one has to call `.with_default_capacity()` if they really don't want to provide or configure the address.

Please let me know how to navigate the upgrade process. I only bumped the Cargo version but we'll also have to publish this, possibly prepare a changelog, etc.

### DIFF
--- a/benches/pub_sub.rs
+++ b/benches/pub_sub.rs
@@ -4,7 +4,7 @@ use std::time::{Duration, Instant};
 use tonari_actor::{Actor, Context, Event, Recipient, System};
 
 #[derive(Debug, Clone)]
-struct StringEvent(String);
+struct StringEvent(());
 
 impl Event for StringEvent {}
 
@@ -45,7 +45,7 @@ impl Actor for PublisherActor {
             PublisherMessage::PublishEvents => {
                 let start = Instant::now();
                 for _i in 0..self.iterations {
-                    context.system_handle.publish(StringEvent("hello".to_string()))?;
+                    context.system_handle.publish(StringEvent(()))?;
                 }
                 let elapsed = start.elapsed();
 
--- a/benches/pub_sub.rs
+++ b/benches/pub_sub.rs
@@ -4,7 +4,7 @@ use std::time::{Duration, Instant};
 use tonari_actor::{Actor, Context, Event, Recipient, System};
 
 #[derive(Debug, Clone)]
-struct StringEvent(());
+struct StringEvent;
 
 impl Event for StringEvent {}
 
@@ -45,7 +45,7 @@ impl Actor for PublisherActor {
             PublisherMessage::PublishEvents => {
                 let start = Instant::now();
                 for _i in 0..self.iterations {
-                    context.system_handle.publish(StringEvent(()))?;
+                    context.system_handle.publish(StringEvent)?;
                 }
                 let elapsed = start.elapsed();
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1035,6 +1035,7 @@ impl<M: Into<N>, N> SenderTrait<M> for Arc<dyn SenderTrait<N>> {
 #[cfg(test)]
 mod tests {
     use std::{
+        rc::Rc,
         sync::atomic::{AtomicU32, Ordering},
         time::Duration,
     };
@@ -1100,7 +1101,9 @@ mod tests {
     #[test]
     fn send_constraints() {
         #[derive(Default)]
-        struct LocalActor;
+        struct LocalActor {
+            _ensure_not_send_not_sync: Rc<()>,
+        }
         impl Actor for LocalActor {
             type Context = Context<Self::Message>;
             type Error = ();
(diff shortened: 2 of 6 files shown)


src/lib.rs
```diff
@@ -319,59 +319,68 @@ impl From<usize> for Capacity {
...
-/// A builder for specifying how to spawn an [`Actor`].
-/// You can specify your own [`Addr`] for the Actor,
-/// the capacity of the Actor's inbox, and you can specify
-/// whether to spawn the Actor into its own thread or block
-/// on the current calling thread.
-#[must_use = "You must call .spawn() or .block_on() to run this actor"]
-pub struct SpawnBuilder<'a, A: Actor, F: FnOnce() -> A> {
+/// A builder for configuring [`Actor`] spawning.
+/// You can specify your own [`Addr`] for the Actor, or let the system create
+/// a new address either provided or default capacity.
```
strohel, line 324:
```suggestion
/// a new address with either provided or default capacity.
```

src/lib.rs
```diff
@@ -319,59 +319,68 @@ impl From<usize> for Capacity {
...
-    pub fn with_addr(self, addr: Addr<A>) -> Self {
-        Self { addr: Some(addr), ..self }
+    // Use the default capacity for the actor's receiving channel.
+    pub fn with_default_capacity(self) -> SpawnBuilderWithAddress<'a, A, F> {
+        let addr = Addr::with_capacity(Capacity::default());
+        SpawnBuilderWithAddress { spawn_builder: self, addr }
     }
+}
 
+#[must_use = "You must call .spawn() or .run_and_block() to create an actor"]
```
strohel, line 352:
```suggestion
#[must_use = "You must call .spawn() or .run_and_block() to run an actor"]
```

src/lib.rs
```diff
@@ -319,59 +319,68 @@ impl From<usize> for Capacity {
...
-        Self { capacity: capacity.into(), ..self }
+    pub fn with_capacity(self, capacity: impl Into<Capacity>) -> SpawnBuilderWithAddress<'a, A, F> {
+        let addr = Addr::with_capacity(capacity);
+        SpawnBuilderWithAddress { spawn_builder: self, addr }
     }
 
-    /// Specify an existing [`Addr`] to use with this Actor.
-    pub fn with_addr(self, addr: Addr<A>) -> Self {
-        Self { addr: Some(addr), ..self }
+    // Use the default capacity for the actor's receiving channel.
```
strohel, line 345:
```suggestion
    /// Use the default capacity for the actor's receiving channel.
```

strohel (commented):
Nice! I have just some little docstrings suggestions.

Can you please pre-test this by using this branch in `portal`, and opening the required changes as a draft PR? To see the effect. Or maybe this doesn't require any `portal` changes?

> Please let me know how to navigate the upgrade process. I only bumped the Cargo version but we'll also have to publish this, possibly prepare a changelog, etc.

I usually just use `cargo-release`. Let's pair on releasing the version once this is merged.

mbernat:
> Nice! I have just some little docstrings suggestions.

Thanks, looking good, I applied all the suggestions.

> Can you please pre-test this by using this branch in `portal`, and opening the required changes as a draft PR? To see the effect. Or maybe this doesn't require any `portal` changes?

It does require changes, here's the PR: https://github.com/tonarino/portal/pull/3333

> I usually just use `cargo-release`. Let's pair on releasing the version once this is merged. Bumping the version in this PR is correct IMO (otherwise we could forget this is semver-breaking change).

Nice, let's do that.

benches/pub_sub.rs
```diff
@@ -4,7 +4,7 @@ use std::time::{Duration, Instant};
 use tonari_actor::{Actor, Context, Event, Recipient, System};
 
 #[derive(Debug, Clone)]
-struct StringEvent(String);
+struct StringEvent(());
```
bschwind, line 7: Was this change due to a clippy warning? If so and we just want a unit struct, then maybe just `struct ActorEvent;` is enough. A tuple struct with just the unit type seems a bit strange.

bschwind (approved):
LGTM! The changes make sense to me, I just have a small comment on one of the examples.

benches/pub_sub.rs
mbernat: Yeah, it's because of clippy. Will fix.

src/lib.rs
```diff
@@ -1095,7 +1100,7 @@ mod tests {
     #[test]
     fn send_constraints() {
         #[derive(Default)]
-        struct LocalActor(Rc<()>);
+        struct LocalActor;
```
strohel, line 1103: Ah sorry, this should actually be:
```suggestion
        struct LocalActor {
            _ensure_not_send_not_sync: Rc<()>,
        }
```
...the intention is that this tests tonari_actor still provides an API to spawn actors that are neither Send nor Sync.

I had this in a local `fix-clippy` branch for a week already, but forgot to push/open PR. :shrug:

strohel (approved):
LGTM! Just one unrelated change request for the fix clippy part. Conditionally approving already.

mbernat:
@strohel Please let me know when you can spare a minute to release this with me, thanks!

Issue #86, opened 2024-08-05
Add cache for the last published event


goodhoko:
Provide a way for subscribers to receive the last published event (if there is such).

PR #87, opened 2024-08-07
Cache last published event

goodhoko:
and let subscribers receive it upon subscription.

Optionally, we could provide a separate marker trait for cached events and persist only those. I couldn't figure out how to dynamically query for the type (trait) in `publish`. Somehow `downcast_ref` is not implemented for `Event` (despite `Event` being `Any`). I'd appreciate some help with that.

This causes performance regression for the pub-sub system. Here's output of `cargo bench --bench pub_sub` running on this branch (comparing it to the previous run on `main`):

<details><summary>Details</summary>
<p>

```
   Compiling tonari-actor v0.9.0 (/Users/jentak/workspace/tonari/actor)
    Finished `bench` profile [optimized] target(s) in 1.20s
     Running benches/pub_sub.rs (target/release/deps/pub_sub-9cceefaff682ce83)
publish/1 publishers 10 subscribers
                        time:   [2.3908 µs 2.8514 µs 3.4074 µs]
                        thrpt:  [293.48 Kelem/s 350.71 Kelem/s 418.27 Kelem/s]
                 change:
                        time:   [+23.719% +56.164% +97.912%] (p = 0.00 < 0.05)
                        thrpt:  [-49.473% -35.965% -19.171%]
                        Performance has regressed.
Found 6 outliers among 100 measurements (6.00%)
  4 (4.00%) high mild
  2 (2.00%) high severe
publish/2 publishers 10 subscribers
                        time:   [3.5475 µs 3.9570 µs 4.4083 µs]
                        thrpt:  [226.84 Kelem/s 252.71 Kelem/s 281.89 Kelem/s]
                 change:
                        time:   [+95.062% +134.00% +177.48%] (p = 0.00 < 0.05)
                        thrpt:  [-63.962% -57.265% -48.734%]
                        Performance has regressed.
Found 6 outliers among 100 measurements (6.00%)
  2 (2.00%) high mild
  4 (4.00%) high severe
publish/10 publishers 10 subscribers
                        time:   [27.460 µs 30.042 µs 32.920 µs]
                        thrpt:  [30.377 Kelem/s 33.287 Kelem/s 36.416 Kelem/s]
                 change:
                        time:   [+84.814% +102.23% +121.67%] (p = 0.00 < 0.05)
                        thrpt:  [-54.888% -50.552% -45.891%]
                        Performance has regressed.
  3 (3.00%) high mild
publish/1 publishers 100 subscribers
                        time:   [75.190 µs 81.870 µs 87.810 µs]
                        thrpt:  [11.388 Kelem/s 12.214 Kelem/s 13.300 Kelem/s]
                 change:
                        time:   [-10.861% -5.4355% -0.4260%] (p = 0.03 < 0.05)
                        thrpt:  [+0.4278% +5.7480% +12.184%]
                        Change within noise threshold.
Found 11 outliers among 100 measurements (11.00%)
  8 (8.00%) low severe
  1 (1.00%) low mild
  2 (2.00%) high severe
publish/2 publishers 100 subscribers
                        time:   [79.105 µs 97.231 µs 115.16 µs]
                        thrpt:  [8.6836 Kelem/s 10.285 Kelem/s 12.641 Kelem/s]
                 change:
                        time:   [-15.101% -0.2671% +14.740%] (p = 0.97 > 0.05)
                        thrpt:  [-12.846% +0.2678% +17.787%]
                        No change in performance detected.
publish/10 publishers 100 subscribers
                        time:   [774.46 µs 781.99 µs 788.91 µs]
                        thrpt:  [1.2676 Kelem/s 1.2788 Kelem/s 1.2912 Kelem/s]
                 change:
                        time:   [+299.05% +448.33% +691.39%] (p = 0.00 < 0.05)
                        thrpt:  [-87.364% -81.763% -74.941%]
                        Performance has regressed.
Found 6 outliers among 100 measurements (6.00%)
  1 (1.00%) low severe
  5 (5.00%) low mild
```

</p>
</details> 

That's likely caused by the contention on `SystemHandle::event_subscribers` RwLock. Previously it only needed exclusive access when subscribing and shared access when publishing. Now it needs exclusive access when subscribing *and* publishing.

Closes https://github.com/tonarino/actor/issues/86

This is a semver-breaking change (due to the new bounds on Event).

### DIFF
--- a/benches/pub_sub.rs
+++ b/benches/pub_sub.rs
@@ -1,10 +1,17 @@
 use anyhow::Error;
 use criterion::{criterion_group, criterion_main, Criterion};
-use std::time::{Duration, Instant};
+use std::{
+    hint::black_box,
+    time::{Duration, Instant},
+};
 use tonari_actor::{Actor, Context, Event, Recipient, System};
 
+const PAYLOAD: &str = "This is the payload that will be used in the test event. It should be of \
+                       reasonable size for a representative event, which is hard to determine. \
+                       But let's say 3 lines of text is fine for now.";
+
 #[derive(Debug, Clone)]
-struct StringEvent;
+struct StringEvent(String);
 
 impl Event for StringEvent {}
 
@@ -45,7 +52,7 @@ impl Actor for PublisherActor {
             PublisherMessage::PublishEvents => {
                 let start = Instant::now();
                 for _i in 0..self.iterations {
-                    context.system_handle.publish(StringEvent)?;
+                    context.system_handle.publish(StringEvent(PAYLOAD.to_string()))?;
                 }
                 let elapsed = start.elapsed();
 
@@ -86,8 +93,10 @@ impl Actor for SubscriberActor {
     fn handle(
         &mut self,
         _context: &mut Self::Context,
-        _message: Self::Message,
+        message: Self::Message,
     ) -> Result<(), Self::Error> {
+        // This black_box has a nice side effect that it silences the 'field is never read' warning.
+        black_box(message.0);
         Ok(())
     }
 }
--- a/benches/pub_sub.rs
+++ b/benches/pub_sub.rs
@@ -4,7 +4,7 @@ use std::{
     hint::black_box,
     time::{Duration, Instant},
 };
-use tonari_actor::{Actor, CacheableEvent, Context, Event, Recipient, System};
+use tonari_actor::{Actor, Context, Event, Recipient, System};
 
 const PAYLOAD: &str = "This is the payload that will be used in the test event. It should be of \
                        reasonable size for a representative event, which is hard to determine. \
@@ -12,11 +12,8 @@ const PAYLOAD: &str = "This is the payload that will be used in the test event. 
 
 #[derive(Debug, Clone)]
 struct StringEvent(String);
-impl Event for StringEvent {}
 
-#[derive(Debug, Clone)]
-struct CacheableStringEvent(String);
-impl CacheableEvent for CacheableStringEvent {}
+impl Event for StringEvent {}
 
 enum PublisherMessage {
     SubscriberStarted,
@@ -24,14 +21,13 @@ enum PublisherMessage {
 }
 
 #[derive(Clone)]
-struct PublisherActor<E> {
-    event: E,
+struct PublisherActor {
     subscriber_count: usize,
     iterations: u64,
     result_sender: flume::Sender<Duration>,
 }
 
-impl<E: Event> Actor for PublisherActor<E> {
+impl Actor for PublisherActor {
     type Context = Context<Self::Message>;
     type Error = Error;
     type Message = PublisherMessage;
@@ -56,7 +52,7 @@ impl<E: Event> Actor for PublisherActor<E> {
             PublisherMessage::PublishEvents => {
                 let start = Instant::now();
                 for _i in 0..self.iterations {
-                    context.system_handle.publish(self.event.clone())?;
+                    context.system_handle.publish(StringEvent(PAYLOAD.to_string()))?;
                 }
                 let elapsed = start.elapsed();
 
@@ -81,7 +77,7 @@ impl SubscriberActor {
 impl Actor for SubscriberActor {
     type Context = Context<Self::Message>;
     type Error = Error;
-    type Message = EitherEvent;
+    type Message = StringEvent;
 
     fn name() -> &'static str {
         "SubscriberActor"
@@ -89,7 +85,6 @@ impl Actor for SubscriberActor {
 
     fn started(&mut self, context: &mut Self::Context) {
         context.subscribe::<StringEvent>();
-        context.subscribe::<CacheableStringEvent>();
         for publisher_addr in self.publisher_addrs.iter() {
             publisher_addr.send(PublisherMessage::SubscriberStarted).unwrap();
         }
@@ -101,43 +96,12 @@ impl Actor for SubscriberActor {
         message: Self::Message,
     ) -> Result<(), Self::Error> {
         // This black_box has a nice side effect that it silences the 'field is never read' warning.
-        black_box(message.payload());
+        black_box(message.0);
         Ok(())
     }
 }
 
-enum EitherEvent {
-    Base(StringEvent),
-    Cacheable(CacheableStringEvent),
-}
-
-impl EitherEvent {
-    fn payload(self) -> String {
-        match self {
-            EitherEvent::Base(StringEvent(payload))
-            | EitherEvent::Cacheable(CacheableStringEvent(payload)) => payload,
-        }
-    }
-}
-
-impl From<StringEvent> for EitherEvent {
-    fn from(value: StringEvent) -> Self {
-        Self::Base(value)
-    }
-}
-
-impl From<CacheableStringEvent> for EitherEvent {
-    fn from(value: CacheableStringEvent) -> Self {
-        Self::Cacheable(value)
-    }
-}
-
-fn run_pubsub_system<E: Event + Send>(
-    event: E,
-    publishers: usize,
-    subscribers: usize,
-    iterations: u64,
-) -> Duration {
+fn run_pubsub_system(publishers: usize, subscribers: usize, iterations: u64) -> Duration {
     let mut system = System::new("pub sub bench");
     let (result_sender, result_receiver) = flume::bounded(publishers);
 
@@ -145,7 +109,6 @@ fn run_pubsub_system<E: Event + Send>(
     let per_publisher_iterations = iterations / publishers as u64;
     let publisher_actors = vec![
         PublisherActor {
-            event,
             subscriber_count: subscribers,
             iterations: per_publisher_iterations,
             result_sender
@@ -187,33 +150,11 @@ fn criterion_benchmark(c: &mut Criterion) {
                 format!("{publishers} publishers {subscribers} subscribers"),
                 |b| {
                     b.iter_custom(|iterations| {
-                        run_pubsub_system(
-                            StringEvent(PAYLOAD.to_string()),
-                            publishers,
-                            subscribers,
-                            iterations,
-                        )
+                        run_pubsub_system(publishers, subscribers, iterations)
                     })
                 },
             );
         }
-    }
-
-    let subscribers = 10;
-    for publishers in [1, 2] {
-        group.bench_function(
-            format!("{publishers} publishers {subscribers} subscribers (cacheable event type)"),
-            |b| {
-                b.iter_custom(|iterations| {
-                    run_pubsub_system(
-                        CacheableStringEvent(PAYLOAD.to_string()),
-                        publishers,
-                        subscribers,
-                        iterations,
-                    )
-                })
-            },
-        );
     }
 }
 
--- a/benches/pub_sub.rs
+++ b/benches/pub_sub.rs
@@ -4,7 +4,7 @@ use std::{
     hint::black_box,
     time::{Duration, Instant},
 };
-use tonari_actor::{Actor, Context, Event, Recipient, System};
+use tonari_actor::{Actor, CacheableEvent, Context, Event, Recipient, System};
 
 const PAYLOAD: &str = "This is the payload that will be used in the test event. It should be of \
                        reasonable size for a representative event, which is hard to determine. \
@@ -12,8 +12,11 @@ const PAYLOAD: &str = "This is the payload that will be used in the test event. 
 
 #[derive(Debug, Clone)]
 struct StringEvent(String);
-
 impl Event for StringEvent {}
+
+#[derive(Debug, Clone)]
+struct CacheableStringEvent(String);
+impl CacheableEvent for CacheableStringEvent {}
 
 enum PublisherMessage {
     SubscriberStarted,
@@ -21,13 +24,14 @@ enum PublisherMessage {
 }
 
 #[derive(Clone)]
-struct PublisherActor {
+struct PublisherActor<E> {
+    event: E,
     subscriber_count: usize,
     iterations: u64,
     result_sender: flume::Sender<Duration>,
 }
 
-impl Actor for PublisherActor {
+impl<E: Event> Actor for PublisherActor<E> {
     type Context = Context<Self::Message>;
     type Error = Error;
     type Message = PublisherMessage;
@@ -52,7 +56,7 @@ impl Actor for PublisherActor {
             PublisherMessage::PublishEvents => {
                 let start = Instant::now();
                 for _i in 0..self.iterations {
-                    context.system_handle.publish(StringEvent(PAYLOAD.to_string()))?;
+                    context.system_handle.publish(self.event.clone())?;
                 }
                 let elapsed = start.elapsed();
 
@@ -77,7 +81,7 @@ impl SubscriberActor {
 impl Actor for SubscriberActor {
     type Context = Context<Self::Message>;
     type Error = Error;
-    type Message = StringEvent;
+    type Message = EitherEvent;
 
     fn name() -> &'static str {
         "SubscriberActor"
@@ -85,6 +89,7 @@ impl Actor for SubscriberActor {
 
     fn started(&mut self, context: &mut Self::Context) {
         context.subscribe::<StringEvent>();
+        context.subscribe::<CacheableStringEvent>();
         for publisher_addr in self.publisher_addrs.iter() {
             publisher_addr.send(PublisherMessage::SubscriberStarted).unwrap();
         }
@@ -96,12 +101,43 @@ impl Actor for SubscriberActor {
         message: Self::Message,
     ) -> Result<(), Self::Error> {
         // This black_box has a nice side effect that it silences the 'field is never read' warning.
-        black_box(message.0);
+        black_box(message.payload());
         Ok(())
     }
 }
 
-fn run_pubsub_system(publishers: usize, subscribers: usize, iterations: u64) -> Duration {
+enum EitherEvent {
+    Base(StringEvent),
+    Cacheable(CacheableStringEvent),
+}
+
+impl EitherEvent {
+    fn payload(self) -> String {
+        match self {
+            EitherEvent::Base(StringEvent(payload))
+            | EitherEvent::Cacheable(CacheableStringEvent(payload)) => payload,
+        }
+    }
+}
+
+impl From<StringEvent> for EitherEvent {
+    fn from(value: StringEvent) -> Self {
+        Self::Base(value)
+    }
+}
+
+impl From<CacheableStringEvent> for EitherEvent {
+    fn from(value: CacheableStringEvent) -> Self {
+        Self::Cacheable(value)
+    }
+}
+
+fn run_pubsub_system<E: Event + Send>(
+    event: E,
+    publishers: usize,
+    subscribers: usize,
+    iterations: u64,
+) -> Duration {
     let mut system = System::new("pub sub bench");
     let (result_sender, result_receiver) = flume::bounded(publishers);
 
@@ -109,6 +145,7 @@ fn run_pubsub_system(publishers: usize, subscribers: usize, iterations: u64) -> 
     let per_publisher_iterations = iterations / publishers as u64;
     let publisher_actors = vec![
         PublisherActor {
+            event,
             subscriber_count: subscribers,
             iterations: per_publisher_iterations,
             result_sender
@@ -150,11 +187,33 @@ fn criterion_benchmark(c: &mut Criterion) {
                 format!("{publishers} publishers {subscribers} subscribers"),
                 |b| {
                     b.iter_custom(|iterations| {
-                        run_pubsub_system(publishers, subscribers, iterations)
+                        run_pubsub_system(
+                            StringEvent(PAYLOAD.to_string()),
+                            publishers,
+                            subscribers,
+                            iterations,
+                        )
                     })
                 },
             );
         }
+    }
+
+    let subscribers = 10;
+    for publishers in [1, 2] {
+        group.bench_function(
+            format!("{publishers} publishers {subscribers} subscribers (cacheable event type)"),
+            |b| {
+                b.iter_custom(|iterations| {
+                    run_pubsub_system(
+                        CacheableStringEvent(PAYLOAD.to_string()),
+                        publishers,
+                        subscribers,
+                        iterations,
+                    )
+                })
+            },
+        );
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -62,7 +62,6 @@ use log::*;
 use parking_lot::{Mutex, RwLock};
 use std::{
     any::TypeId,
-    collections::HashMap,
     fmt,
     ops::Deref,
     sync::Arc,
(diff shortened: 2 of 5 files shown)


strohel:
I've pushed 4 commits to this, @goodhoko (and others) PTAL.

It makes sense to review them one-by-one (and check their messages) if you're curious.

It would be also interesting to run `cargo bench --bench pub_sub` on each commit starting from `main` till the current tip of this branch.

src/lib.rs
```diff
@@ -772,16 +793,18 @@ impl SystemHandle {
 
     /// Publish an event. All actors that have previously subscribed to the type will receive it.
     ///
-    /// The event will be also cached. Actors that will subscribe to the type in future may choose
-    /// to receive the last cached event upon subscription.
+    /// The event will be cached if it implements [`CacheableEvent`]. Actors that will subscribe
+    /// such an event type in future may choose to receive the last cached event upon subscription.
```
goodhoko, line 784:
```suggestion
    /// The event will be cached if it implements [`CacheableEvent`]. Actors that will subscribe to
    /// such an event type in future may choose to receive the last cached event upon subscription.
```

src/lib.rs
```diff
@@ -225,13 +225,34 @@ enum SystemState {
...
+/// of the actor framework and can be cached. [`Context::subscribe_and_receive_latest()`]  will be
+/// available for them in addition to [`Context::subscribe()`]. Blanket implementation of [`Event`]
+/// is provided for all types that implement [`CacheableEvent`].
+pub trait CacheableEvent: Clone + std::any::Any + Send + Sync {}
+
+impl<T: CacheableEvent> Event for T {
+    fn cache(&self) -> Option<Box<dyn std::any::Any + Send + Sync>> {
+        Some(Box::new(self.clone()))
+    }
+}
```
goodhoko, line 249: Nice!

src/lib.rs
```diff
@@ -225,13 +225,34 @@ enum SystemState {
...
+    fn cache(&self) -> Option<Box<dyn std::any::Any + Send + Sync>> {
+        None
+    }
+}
+
+/// A marker trait for types which participate in the publish-subscribe system
+/// of the actor framework and can be cached. [`Context::subscribe_and_receive_latest()`]  will be
+/// available for them in addition to [`Context::subscribe()`]. Blanket implementation of [`Event`]
+/// is provided for all types that implement [`CacheableEvent`].
+pub trait CacheableEvent: Clone + std::any::Any + Send + Sync {}
```
goodhoko, line 243: I think this should be called `CachedEvent` because if you use it, we don't ask you "would you like to cache this event, please?" - we just cache it.

src/lib.rs
```diff
@@ -225,13 +225,34 @@ enum SystemState {
...
-pub trait Event: Clone + std::any::Any + 'static + Send + Sync {}
+/// of the actor framework. If it makes sense to cache your event and use it with
+/// [`Context::subscribe_and_receive_latest()`], then mark your type with [`CacheableEvent`]
+/// instead.
+pub trait Event: Clone + std::any::Any {
+    /// Internal method for the actor framework: don't implement this downstream. Make your type
+    /// implement [`CacheableEvent`] instead of [`Event`] if it makes sense to cache it.
+    #[doc(hidden)]
+    fn cache(&self) -> Option<Box<dyn std::any::Any + Send + Sync>> {
+        None
```
goodhoko, line 235: In theory, letting users implement this would make it possible to condition caching on their own logic. Imagine, us saying that we want to cache only some instances of `PortalState`.

But let's keep this "private" unless we find a compelling usecase for it.

src/lib.rs
```diff
@@ -766,28 +763,18 @@ impl SystemHandle {
         &self,
         recipient: Recipient<M>,
     ) -> Result<(), SendError> {
-        let mut event_subscribers = self.event_subscribers.write();
+        // Subscribe first. We assume it is better to receive an event twice rather than not at all.
```
goodhoko, line 766: This is weird assumption for me. With the previous logic, we can only miss an event if it is published right in between us checking the cache and adding the callback to event_subscribers. That is a very short window and at that point the notion of "last published event" is unclear: from POV of the user, that missed event was published *after* they called `subscribe_and_receive_latest()` so it's expected the subscriber won't receive it.

We can't assume it's ok to deliver events twice - that depends on the semantics of the subscriber, not the event itself.

I think we should send cached first and subscribe later.

goodhoko (commented):
@strohel Nice, thanks! This looks good but I have some functional remarks and questions.

src/lib.rs
strohel: > In theory, letting users implement this would make it possible to condition caching on their own logic.

Well, not with the current design: subsribers can only `subscribe_and_receive_latest()` to `CacheableEvent`s, and it is not possible for downstream crates to implement `Event` for anything that already implements `CacheableEvent` (it would conflict with the blanket implementation that we provide).

src/lib.rs
goodhoko: discussed in person: Cacheable is fine!

strohel:
@goodhoko @bschwind alright I've:
1. prepended a commit that makes the pub_sub benchmark a bit more realistic (previously the payload was zero-sized-type)
2. Measured the commit-by-commit performance on P3 Tiny (tonari services stopped) using:
   `git rebase main --exec 'cargo bench --bench pub_sub -- --save-baseline "$(git describe | tr "\n" " "; git show -q --format=%s | cut -d" " -f -4)"'`
3. Tabulated the results using `critcmp --list`:

```
publish/1 publishers 10 subscribers
-----------------------------------
v0.9.0-1-gd3713d6 Make pub_sub bench more                1.05      10.7±2.83µs  91.5 KElem/sec
v0.9.0-2-gb53014d Cache last published event             1.00      10.2±1.52µs  95.3 KElem/sec
v0.9.0-3-g508a62d Try to use dashmap                     1.01      10.3±1.80µs  94.6 KElem/sec
v0.9.0-4-gc3a0d56 Split `Event` and `CacheableEvent`     1.00      10.2±1.56µs  95.7 KElem/sec
v0.9.0-5-g2485610 Extend pub_sub benchmark with          1.02      10.4±1.81µs  93.7 KElem/sec
v0.9.0-6-gb5d4d3f Use DashMap also for                   1.06      10.8±2.13µs  90.6 KElem/sec
v0.9.0-7-g171a9c7 Drop `dashmap` and reinstate           1.03      10.5±1.39µs  93.1 KElem/sec
v0.9.0-8-g148227a Deduplicate subscription code into     1.02      10.4±1.48µs  93.7 KElem/sec

publish/1 publishers 10 subscribers (cacheable event type)
----------------------------------------------------------
v0.9.0-5-g2485610 Extend pub_sub benchmark with          1.06      10.6±1.84µs  92.1 KElem/sec
v0.9.0-6-gb5d4d3f Use DashMap also for                   1.00      10.0±1.22µs  97.4 KElem/sec
v0.9.0-7-g171a9c7 Drop `dashmap` and reinstate           1.08      10.9±2.65µs  89.9 KElem/sec
v0.9.0-8-g148227a Deduplicate subscription code into     1.09      10.9±2.08µs  89.5 KElem/sec

publish/2 publishers 10 subscribers
-----------------------------------
v0.9.0-1-gd3713d6 Make pub_sub bench more                1.00       9.4±0.53µs  103.8 KElem/sec
v0.9.0-2-gb53014d Cache last published event             2.81      26.4±7.82µs  36.9 KElem/sec
v0.9.0-3-g508a62d Try to use dashmap                     1.05       9.8±0.62µs  99.3 KElem/sec
v0.9.0-4-gc3a0d56 Split `Event` and `CacheableEvent`     1.02       9.6±0.46µs  102.1 KElem/sec
v0.9.0-5-g2485610 Extend pub_sub benchmark with          1.01       9.5±0.48µs  103.0 KElem/sec
v0.9.0-6-gb5d4d3f Use DashMap also for                   1.00       9.4±0.46µs  103.7 KElem/sec
v0.9.0-7-g171a9c7 Drop `dashmap` and reinstate           1.02       9.6±0.46µs  101.6 KElem/sec
v0.9.0-8-g148227a Deduplicate subscription code into     1.00       9.4±0.43µs  103.5 KElem/sec

publish/2 publishers 10 subscribers (cacheable event type)
----------------------------------------------------------
v0.9.0-5-g2485610 Extend pub_sub benchmark with          1.00       9.6±0.42µs  101.5 KElem/sec
v0.9.0-6-gb5d4d3f Use DashMap also for                   1.00       9.7±0.50µs  101.2 KElem/sec
v0.9.0-7-g171a9c7 Drop `dashmap` and reinstate           2.87      27.6±5.70µs  35.4 KElem/sec
v0.9.0-8-g148227a Deduplicate subscription code into     2.71      26.1±3.50µs  37.4 KElem/sec

publish/10 publishers 10 subscribers
------------------------------------
v0.9.0-1-gd3713d6 Make pub_sub bench more                1.00       5.2±1.10µs  188.6 KElem/sec
v0.9.0-2-gb53014d Cache last published event            20.98    108.6±25.14µs  9.0 KElem/sec
v0.9.0-3-g508a62d Try to use dashmap                     1.09       5.6±1.00µs  173.5 KElem/sec
v0.9.0-4-gc3a0d56 Split `Event` and `CacheableEvent`     1.02       5.3±1.16µs  184.2 KElem/sec
v0.9.0-5-g2485610 Extend pub_sub benchmark with          1.24       6.4±3.09µs  151.6 KElem/sec
v0.9.0-6-gb5d4d3f Use DashMap also for                   1.06       5.5±1.15µs  177.6 KElem/sec
v0.9.0-7-g171a9c7 Drop `dashmap` and reinstate           1.04       5.4±1.12µs  180.9 KElem/sec
v0.9.0-8-g148227a Deduplicate subscription code into     1.07       5.6±1.42µs  175.6 KElem/sec

publish/10 publishers 100 subscribers
-------------------------------------
v0.9.0-1-gd3713d6 Make pub_sub bench more                1.00    273.8±23.13µs  3.6 KElem/sec
v0.9.0-2-gb53014d Cache last published event             8.13       2.2±0.66ms  449 Elem/sec
v0.9.0-3-g508a62d Try to use dashmap                     1.04    283.6±36.83µs  3.4 KElem/sec
v0.9.0-4-gc3a0d56 Split `Event` and `CacheableEvent`     1.04    283.5±34.94µs  3.4 KElem/sec
v0.9.0-5-g2485610 Extend pub_sub benchmark with          1.05    286.1±34.07µs  3.4 KElem/sec
v0.9.0-6-gb5d4d3f Use DashMap also for                   1.06    289.5±52.77µs  3.4 KElem/sec
v0.9.0-7-g171a9c7 Drop `dashmap` and reinstate           1.06    289.2±45.56µs  3.4 KElem/sec
v0.9.0-8-g148227a Deduplicate subscription code into     1.03    281.5±28.75µs  3.5 KElem/sec

publish/1 publishers 100 subscribers
------------------------------------
v0.9.0-1-gd3713d6 Make pub_sub bench more                1.05    125.9±44.44µs  7.8 KElem/sec
v0.9.0-2-gb53014d Cache last published event             1.04    124.7±55.50µs  7.8 KElem/sec
v0.9.0-3-g508a62d Try to use dashmap                     1.05    125.6±50.27µs  7.8 KElem/sec
v0.9.0-4-gc3a0d56 Split `Event` and `CacheableEvent`     1.01    121.4±34.90µs  8.0 KElem/sec
v0.9.0-5-g2485610 Extend pub_sub benchmark with          1.13    134.8±53.37µs  7.2 KElem/sec
v0.9.0-6-gb5d4d3f Use DashMap also for                   1.06    127.3±42.72µs  7.7 KElem/sec
v0.9.0-7-g171a9c7 Drop `dashmap` and reinstate           1.00    119.8±32.54µs  8.2 KElem/sec
v0.9.0-8-g148227a Deduplicate subscription code into     1.03    123.1±31.26µs  7.9 KElem/sec

publish/2 publishers 100 subscribers
------------------------------------
v0.9.0-1-gd3713d6 Make pub_sub bench more                1.08    120.6±26.07µs  8.1 KElem/sec
v0.9.0-2-gb53014d Cache last published event             4.17   466.3±150.09µs  2.1 KElem/sec
v0.9.0-3-g508a62d Try to use dashmap                     1.00    111.8±10.41µs  8.7 KElem/sec
v0.9.0-4-gc3a0d56 Split `Event` and `CacheableEvent`     1.03    115.6±17.81µs  8.4 KElem/sec
v0.9.0-5-g2485610 Extend pub_sub benchmark with          1.01     112.6±7.52µs  8.7 KElem/sec
v0.9.0-6-gb5d4d3f Use DashMap also for                   1.08    120.3±20.32µs  8.1 KElem/sec
v0.9.0-7-g171a9c7 Drop `dashmap` and reinstate           1.07    119.7±20.05µs  8.2 KElem/sec
v0.9.0-8-g148227a Deduplicate subscription code into     1.04    115.8±16.05µs  8.4 KElem/sec
```

src/lib.rs
```diff
@@ -710,26 +765,56 @@ impl SystemHandle {
...
+
+        let event_subscribers = if let Some(cached) = event.cache() {
+            // This write lock is a possible contention point if many actors publish events at high
+            // rate at the same time. If that ever becomes a bottleneck, we could transition to
+            // a hash map with finer locking, like dashmap. Like in anx; intermediate version of #87.
+            let mut event_subscribers = self.event_subscribers.write();
+            event_subscribers.last_value_cache.insert(type_id, cached);
+
+            // Release the write lock early so that the send loop is run only with a read lock,
+            // preventing contention. We make use of the guarantee that no other write lock can ba
```
goodhoko, line 809:
```suggestion
            // preventing contention. We make use of the guarantee that no other write lock can be
```

src/lib.rs
```diff
@@ -710,26 +765,56 @@ impl SystemHandle {
...
-            if let Some(event) = e.downcast_ref::<E>() {
-                let msg = event.clone();
-                recipient.send(msg.into())?;
+        // Send the last cached value if there is one.
+        if let Some(last_cached_value) = event_subscribers.last_value_cache.get(&TypeId::of::<E>())
+        {
+            if let Some(msg) = last_cached_value.downcast_ref::<E>() {
+                recipient.send(msg.clone().into())?;
             }
+        }
```
goodhoko, line 785: I wonder it it would be worth acquiring only a shared access for getting and sending the cached event, and only upgrading to exclusive access for pushing to subscribers.

If we decide to do that we need to be careful not to introduce the possibility of missed events. That depends on how `upgrade()` behaves w.r.t. other writers waiting for exclusive access. It'll work only if `upgrade()` has priority over `write()`. The docs don't seem to mention that so it's likely *not* the case? If not, this could happen

1. `subscribe_and_receive_latest()` acquires a read lock.
2. The cache is empty. No last event to send.
3. `publish()` requests exclusive access. This blocks because `subscribe_and_receive_latest()` has the read lock.
4. `subscribe_and_receive_latest()` tries to `upgrade()`, but `publish()` is already waiting in line and gets exclusive access first.
5. `publish()` saves the event to cache, fans it out to subscribers and releases its lock.
6. `subscribe_and_receive_latest()` is unblocked, finally upgrades and pushes to subscribers. But it's too late.


If we don't do this, we should add a comment explaining why we don't this to prevent future ourselves introducing this subtle bug. Something like:

```
// Require exclusive access for the entire duration of this function to ensure no one
// publishes an event after we checked the cache and before we push to subscribers.
// In such case the event wouldn't be delivered to this subscriber.
```

src/lib.rs
```diff
@@ -710,26 +765,56 @@ impl SystemHandle {
...
+    /// such an event type in future may choose to receive the last cached event upon subscription.
+    ///
     /// When sending to some subscriber fails, others are still tried and vec of errors is returned.
     /// For direct, non-[`Clone`] or high-throughput messages please use [`Addr`] or [`Recipient`].
     pub fn publish<E: Event>(&self, event: E) -> Result<(), PublishError> {
-        let event_subscribers = self.event_subscribers.read();
-        if let Some(subs) = event_subscribers.events.get(&TypeId::of::<E>()) {
+        let type_id = TypeId::of::<E>();
+
+        let event_subscribers = if let Some(cached) = event.cache() {
```
goodhoko, line 801: hehe, nice dance with the variables here!

src/lib.rs
```diff
@@ -710,26 +765,56 @@ impl SystemHandle {
...
+            // This write lock is a possible contention point if many actors publish events at high
+            // rate at the same time. If that ever becomes a bottleneck, we could transition to
+            // a hash map with finer locking, like dashmap. Like in anx; intermediate version of #87.
```
goodhoko, line 804: > Like in anx

what is anx?

> intermediate version of #87.

We should just link the particular commit.

goodhoko (commented):
Nice work, @strohel! Thanks for taking this over, it's a pleasure see you work on something that excites you!

The benchmark results look good! If I read them right, this PR no longer causes noticeable performance regression. This is equally true for using dashmap and intelligent locking (i.e. `upgrade()`).

There's a significant (and expected) difference in performance of the cacheable event when there are multiple publishers between Dashmap and intelligent locking. IMO that's totally ok!

I have this one more thorough read and it looks good, though I have couple last remarks.

src/lib.rs
strohel: The optimization you describe would be possible with https://docs.rs/parking_lot/latest/parking_lot/type.RwLock.html#method.upgradable_read.

But this is _subscribe_: it is called only once per event and subscriber -> not worth optimizing IMO. (I'll also post a general comment on the PR).

src/lib.rs
strohel: anx is a typo, will fix. I plan to change the other part anyway..

strohel:
@bschwind @goodhoko (sorry I meant to submit this right after the benchmark results

## Benchmark Observations
- using `dashmap` for the last event cache _completely_ hides the performance penalty of caching the events
- Even with trying to hold the write lock least amount of time, using RwLock still makes cached messages 3 times slower than non-cached ones in case of 2 subscribers (and probably even slower with more) :-(
- using `dashmap` for the subscriber callback list doesn't have any effect (and we need an RwLock to prevent a race condition anyway)

## Conclusion
- I don't think the `Event` / `CacheableEvent` distinction is worth it, if we can have all the goodies (and and even more performance) by just introducing a dependency
- I'll therefore iterate more to get this to the state around 508a62d `Try to use dashmap`
  - I feel somewhat silly spending so much time for this (I should have _properly_ benchmarked right from start)

src/lib.rs
strohel: (removing the code altogether)

strohel:
> * I'll therefore iterate more to get this to the state around [508a62d](https://github.com/tonarino/actor/commit/508a62d5f8380fd010637ace08c1950688a3bcd0) `Try to use dashmap`, just with the deduplication and nicer comments

- I've pushed this in `Back to dashmap, but keep deduplication and extend comments` ba3dba1 
- The commits have now a lot of back and forth, but the overall change that I've made on top of @goodhoko's initial version is pretty small: https://github.com/tonarino/actor/pull/87/files/b53014d7b45533268baf2dbaa0e3bf99f648b328..ba3dba1225d04105bf8a8037ec9ee1d11f47f595 (this excludes the benchmark change)
- I've also updated the benchmarks in https://github.com/tonarino/actor/pull/87#issuecomment-2276156446 It has the same performance as when we weren't caching at all, in all scenarios (!)

bschwind (approved):
LGTM! No nits or issues with the code that I see :)

goodhoko (commented):
LGTM! (I can't formally approve as the author of this PR.)

Shall we do the release as part of this PR?

src/lib.rs
```diff
@@ -710,26 +746,48 @@ impl SystemHandle {
...
+    ) -> Result<(), SendError> {
+        let mut event_subscribers = self.event_subscribers.write();
 
-        subs.push(Box::new(move |e| {
-            if let Some(event) = e.downcast_ref::<E>() {
-                let msg = event.clone();
-                recipient.send(msg.into())?;
+        // Send the last cached value if there is one. The cached event sending and adding ourselves
+        // to the subscriber list needs to be under the same write lock guard to avoid race
+        // conditions between this method and `publish()` (to guarantee exactly-once delivery).
```
mbernat, line 762: It seems to me that `self.event_subscribers` usage is very correlated with `last_value_cache` usage. Wouldn't it be better to put both the subscribers hashmap and the cache dashmap behind a single RwLock?

benches/pub_sub.rs
```diff
@@ -86,8 +93,10 @@ impl Actor for SubscriberActor {
     fn handle(
         &mut self,
         _context: &mut Self::Context,
-        _message: Self::Message,
+        message: Self::Message,
     ) -> Result<(), Self::Error> {
+        // This black_box has a nice side effect that it silences the 'field is never read' warning.
+        black_box(message.0);
```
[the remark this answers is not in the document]
strohel: This is part of the https://github.com/tonarino/actor/pull/87/commits/d3713d6db110ef91b590667a6c4a5987fa83d570 commit that intentionally changes the performance profile of the benchmark (and does nothing else). We take it as the new baseline.

It is here mainly to suppress the warning, but also to prevent an unlikely situation that this whole method is more or less compiled out.

src/lib.rs
strohel: Well, they are already under a single `RwLock`, in
https://github.com/tonarino/actor/blob/ba3dba1225d04105bf8a8037ec9ee1d11f47f595/src/lib.rs#L263
https://github.com/tonarino/actor/blob/ba3dba1225d04105bf8a8037ec9ee1d11f47f595/src/lib.rs#L232-L239
..? (with slightly misleading variable/struct name)

The motivation for the comments is more that it would be possible to put `last_value_cache` outside of the lock, but we don't want it.

mbernat: OIC, I completely misunderstood the comments, sorry!

mbernat (approved):
LGTM

goodhoko:
This was successfully tested in https://github.com/tonarino/portal/pull/3746. Should be good to merge!

src/lib.rs
```diff
@@ -290,6 +308,19 @@ impl<M> Context<M> {
...
         self.system_handle.subscribe_recipient::<M, E>(self.myself.clone());
     }
+
+    /// Subscribe current actor to event of type `E` and send the last cached event to it.
+    /// This is part of the event system. You don't need to call this method to receive
+    /// direct messages sent using [`Addr`] and [`Recipient`].
+    ///
+    /// Note that subscribing twice to the same event would result in duplicate events -- no
+    /// de-duplication of subscriptions is performed.
+    pub fn subscribe_and_receive_latest<E: Event + Into<M>>(&self) -> Result<(), SendError>
```
strohel, line 318: Clarifying the docstring as per live discussion with @goodhoko.

```suggestion
    /// Subscribe current actor to event of type `E` and send the last cached event to it.
    /// This is part of the event system. You don't need to call this method to receive
    /// direct messages sent using [`Addr`] and [`Recipient`].
    ///
    /// Note that subscribing twice to the same event would result in duplicate events -- no
    /// de-duplication of subscriptions is performed.
    ///
    /// This method may fail if it is not possible to send the latest event. In this case it is
    /// guaranteed that the subscription did not take place. You can safely try again.
    pub fn subscribe_and_receive_latest<E: Event + Into<M>>(&self) -> Result<(), SendError>
```

PR #88, opened 2024-08-12
Error reporting QoL

strohel:
- require Display instead of Debug for Actor::Error associated type
- print handle() (and friends) errors using `{:#}` rather than `{:?}`
  - this hides the backtrace print of anyhow errors
- differentiate handle() (and friends) errors while shutting down vs. running and log them with lower severity
- make it more clear and less spammy when trying to shut down a system while it is already shutting down

This is a semver-breaking change (due to the new bound on Actor::Error).


goodhoko (approved):
LGTM!

PR #89, opened 2024-08-12
Make `Actor::started()`, `stopped()` fallible

strohel:
- on top of #88

I've long time wanted to do this, other work on actor has finally triggered me.

This should allow removing almost all remaining occurrences of fallible `.unwrap()` / `.expect()` in our downstream code.

It is (obviously) semver-breaking change which causes a bit of churn to adapt to, but the fact of adapting itself should be a nice cleanup.

### DIFF
--- a/benches/pub_sub.rs
+++ b/benches/pub_sub.rs
@@ -1,4 +1,4 @@
-use anyhow::Error;
+use anyhow::{Error, Result};
 use criterion::{criterion_group, criterion_main, Criterion};
 use std::{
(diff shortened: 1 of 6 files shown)


mbernat (approved):
Makes sense, neat.

goodhoko (approved):
LGTM!

Let me rebase https://github.com/tonarino/portal/pull/3746 on top of this. It will get rid of quite some lines of `expect`s.

PR #93, opened 2025-07-30
Actorless Addr: transition from `Addr<A: Actor>` to `Addr<M: Message>` (API breaking)

strohel:
- on top of https://github.com/tonarino/actor/pull/97 (not logically needed for this PR, just for linear branch graph)

4.5 years later after my evaluation project for tonari, I've realized there's a slightly better solution to a problem I tried solving in #38 (erasing the actor's type from the `Addr` struct). The `Addr` struct simply doesn't _need_ the `Actor` type parameter, it only needs the message one.

This change causes quite a churn (people need to change their `Addr` generic parameters), the little type design improvement wouldn't be worth it.

But it turns out it is a **prerequisite for supporting `async` actors** (see Q&A, https://github.com/tonarino/actor/pull/94), which pushed it past the worthiness line for me.

The biggest caveat (beyond the API churn) is that `Addr` no longer implements `Default`. It simply no longer has enough information to know the correct capacities of its channels. So instead of `Addr<MyActor>::default()` one needs to write `MyActor::addr()`.

OTOH this allowed some small cleanups especially around `Capacity`, and removal of some `'static` bounds. Notice this is net negative code lines change.

### Why is this a prerequisite of `async` actors?
To support [async actors](https://github.com/tonarino/actor/pull/94), we need to introduce some `AsyncActor` trait, where `handle()` and friends are `async` functions. But `Addr<A>` was using the `A: Actor` bound, which couldn't be lifted, as we need to access the `A::Message` associated type.

There were a couple of options:
1. :thinking: Introduce an `AsyncAddr<A: AsyncActor>` type distinct from `Addr<A: Actor>`
   - likely feasible, but I didn't like the distinction: I hoped the 2 ecosystems could feel more similar and less distinct. Would be somehow mitigated by the fact that `Recipient` could have stayed the same (I think)
2. :thinking: Introduce a `BaseActor` trait that would only define the `Message`, capacities associated type/consts, and have `Actor: BaseActor`, `AsyncActor: BaseActor` that would define `handle()` and friends.
   - this would cause more churn than this PR. _Every_ actor would need 2 impl blocks, each implementing a part of actor functionality. 
3. :heavy_check_mark: Change `Addr<A>` to `Addr<M>` (this PR)
4. :thinking: Any other option that didn't occur to me?

### This blurs the line between `Addr<M>` and `Recipient<M>`. Do we still need `Recipient`?

Yes. While they now have the same type parameter, there's still conceptual difference between `Addr` and `Recipient`:
- `Addr` can (also) be used to _spawn_ actors (receive messages)
- `Recipient` can ony be used to send messages; it doesn't keep the receiving ends of the channels alive

Thanks to this limitation `Recipient` can do the `M` -> `N` message conversion, which is handy. For merely sending messages, it's still recommended to solely use `Recipient`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -406,10 +406,8 @@ impl<A: Actor<Context = Context<<A as Actor>::Message>>, F: FnOnce() -> A>
     }
 }
 
-impl<
-    A: 'static + Actor<Context = Context<<A as Actor>::Message>>,
-    F: FnOnce() -> A + Send + 'static,
-> SpawnBuilderWithAddress<'_, A, F>
+impl<A: Actor<Context = Context<<A as Actor>::Message>>, F: FnOnce() -> A + Send + 'static>
+    SpawnBuilderWithAddress<'_, A, F>
 {
     /// Spawn this Actor into a new thread managed by the [`System`].
     pub fn spawn(self) -> Result<Addr<A::Message>, ActorError> {
@@ -481,7 +479,7 @@ impl System {
     ) -> Result<(), ActorError>
     where
         F: FnOnce() -> A + Send + 'static,
-        A: Actor<Context = Context<<A as Actor>::Message>> + 'static,
+        A: Actor<Context = Context<<A as Actor>::Message>>,
     {
         // Hold the lock until the end of the function to prevent the race
         // condition between spawn and shutdown.


bschwind (approved):
LGTM!

strohel:
> @strohel does this need another review?

I don't think so! I'm just not merging the entire chain until:
- #94 is finished
- portal code is experimentally ported on it and validates it's working

(as without async actors, this prep PRs wouldn't be worth the churn).

PR #94, opened 2025-07-30
WIP: Add support for `async` actors

strohel:
- on top of https://github.com/tonarino/actor/pull/93 (hard prerequisite of this type system design)

WIP support for async actors. Already works, but still work to be done:

## TODOs
- [ ] enable all features on CI
- [ ] mimic sync actor spawn builders in the async module
- [ ] finish module-level documentation of `async.rs` - particularly document behavior of `async fn handle()` and when to spawn a sub-task, tokoo experimental config that needs to be doen, `spawn` + `spawn_local`
- [ ] write an example that reads a file, reaches network, ..
- [ ] write a bench, measure performance (of the actor loop) is comparable with the sync version
- [ ] resolve the big TODO in run_async_actor_select_loop(), re-measure performance


strohel:
This is now ready for reviews! @mbernat / @bschwind / eventually @goodhoko.

src/async.rs
```diff
@@ -0,0 +1,503 @@
...
+use futures_util::{StreamExt, select_biased};
+use log::{debug, trace};
+use std::{any::type_name, fmt, future, thread};
+use tokio::runtime::LocalRuntime;
+
+/// The actor trait - async variant.
+// Ad. the #[allow]: using `async fn` in a trait doesn't allow us to specify `Send` (or other)
+// bounds, but we don't really need any bounds, because we use [`LocalRuntime`].
+#[allow(async_fn_in_trait)]
+pub trait AsyncActor {
```
mbernat: Could we have an actor super trait to avoid duplicating most of the items? Only the `started()` / `handle()` / `stopped()` methods seem to be different.

examples/async_actors.rs
```diff
@@ -0,0 +1,208 @@
...
+use std::{mem, time::Duration};
+use tokio::{
+    io::{AsyncReadExt, AsyncWriteExt},
+    net::TcpStream,
+    task::JoinHandle,
+    time::{Instant, sleep_until, timeout},
+};
+use tonari_actor::{Actor, BareContext, Context, Recipient, System, r#async::AsyncActor};
+
+enum CollectorTask {
```
mbernat:
```suggestion
enum CrawlerMessage {
```

mbernat (approved):
This is great, not just the impl but also benches, tests and examples.

Looking forward to using async libs in the handlers and replacing a bunch of IO-bound sync actors with a single async one.

benches/benches.rs
```diff
@@ -13,64 +15,145 @@ impl Actor for ChainLink {
...
-        let mut spawn = c.benchmark_group("spawn");
-        spawn.throughput(Throughput::Elements(1));
-        spawn.measurement_time(Duration::from_secs(10));
-        spawn.bench_function("spawn 1", |b| b.iter(|| make_chain(black_box(1))));
-        spawn.bench_function("spawn 10", |b| b.iter(|| make_chain(black_box(10))));
-        spawn.bench_function("spawn 50", |b| b.iter(|| make_chain(black_box(50))));
-    }
+fn spawn_bench(c: &mut Criterion) {
+    // The benchmark may fail at runtime if we i.e. spawn too many async runtimes. The failures
+    // look mysterious without the logs. Logging may be already ini by another bench.
```
bschwind:
```suggestion
    // look mysterious without the logs. Logging may be already initialized by another bench.
```

benches/benches.rs
bschwind: I ran the benches first on `main`, then this branch and got an interesting result:

## main
```
circular/circular (50 actors)
                        time:   [15.668 ms 15.731 ms 15.798 ms]
                        thrpt:  [63.297 Kelem/s 63.567 Kelem/s 63.825 Kelem/s]
  1 (1.00%) low mild
  1 (1.00%) high mild
  1 (1.00%) high severe
```

## this branch

```
circular/circular (50 actors)
                        time:   [1.7004 ms 1.7263 ms 1.7756 ms]
                        thrpt:  [563.20 Kelem/s 579.26 Kelem/s 588.08 Kelem/s]
                 change:
                        time:   [−89.214% −89.026% −88.634%] (p = 0.00 < 0.05)
                        thrpt:  [+779.79% +811.27% +827.10%]
                        Performance has improved.

```

I'm guessing this is just a fundamentally different thing we're benchmarking now?

examples/async_actors.rs
```diff
@@ -0,0 +1,208 @@
+//! A realistic but toy example of using 2 async and 1 sync actor:
+//! - the [`Crawler`] async actor concurrently fetches websites, emits chunks of their data to:
+//! - the [`Sorter`] async actor, which "sorts" the chunks from all websites by delaying each chunk
+//!   according to the value of its first character.
+//! - the [`Collector`] sync actor that prints chunks prefixed by the website as they come.
```
bschwind:
```suggestion
//! - the [`Collector`] sync actor prints chunks prefixed by the website as they come.
```

src/async.rs
```diff
@@ -0,0 +1,503 @@
...
+//! Sync and async actors share the same [`Addr`] and [`Recipient`](crate::Recipient) types.
+//!
+//! Async actors share the same paradigm as sync actors: each one gets its own OS-level thread.
+//! More specifically a single-threaded async runtime is spawned for every async actor.
+//!
+//! `tonari-actor` currently uses the [`tokio`][^tokio] ecosystem, more specifically its
+//! [`LocalRuntime`]. It allows spawning futures that are _not_ [`Send`], which means you
+//! can use [`Rc`](https://doc.rust-lang.org/std/rc/struct.Rc.html) and
+//! [`RefCell`](https://doc.rust-lang.org/std/cell/struct.RefCell.html) instead of
+//! [`Arc`](https://doc.rust-lang.org/std/sync/struct.Arc.html) and mutexes in your futures.
```
bschwind: Just for completeness' sake (I was reading this via `cargo doc`)

```suggestion
//! [`Arc`](https://doc.rust-lang.org/std/sync/struct.Arc.html) and [`Mutex`](https://doc.rust-lang.org/std/sync/struct.Mutex.html)es in your futures.
```

examples/async_actors.rs
```diff
@@ -0,0 +1,208 @@
...
+    type Message = CollectorTask;
+
+    async fn handle(
+        &mut self,
+        _context: &BareContext<CollectorTask>,
+        message: CollectorTask,
+    ) -> Result<()> {
+        match message {
+            CollectorTask::Crawl { hosts } => {
+                // Future combinators are sufficient for concurrent operation within a single actor message.
```
bschwind: If I understand correctly, no more messages can arrive to this actor until this crawling task completes. If that's correct, it might be good to add a note here for better understanding of async actor behavior.

src/lib.rs
```diff
@@ -1348,7 +1356,10 @@ mod tests {
 
     #[test]
     fn message_priorities() {
-        env_logger::Builder::from_env(env_logger::Env::default().default_filter_or("trace")).init();
+        // Logger might have been initialized by another test, so just try on best-effort basis.
```
bschwind:
```suggestion
        // Logger might have been initialized by another test, so just try on a best-effort basis.
```

examples/async_actors.rs
```diff
@@ -0,0 +1,208 @@
...
+use env_logger::Env;
+use futures_util::{StreamExt, TryStreamExt, future::join_all, stream};
+use std::{mem, time::Duration};
+use tokio::{
+    io::{AsyncReadExt, AsyncWriteExt},
+    net::TcpStream,
+    task::JoinHandle,
+    time::{Instant, sleep_until, timeout},
+};
+use tonari_actor::{Actor, BareContext, Context, Recipient, System, r#async::AsyncActor};
```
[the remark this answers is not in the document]
strohel: Yeah good call, will do. I see this boils down only to `AsyncActor`, as the extra methods on `System` are accessible without importing, and `AsyncSpawnBuilder`s also don't need importing as they aren't constructed directly and not usually passed around.

src/async.rs
strohel: That was also my first thought, but I didn't want to be misleading here, because there's [nuance in using `std::sync::Mutex` in async code: you shouldn't hold it locked across .await points](https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html). But maybe I'm overthinking it?

.cargo/config.toml
```diff
@@ -0,0 +1,4 @@
+# See https://docs.rs/tokio/latest/tokio/index.html#unstable-features
+# Downstream crates also need to enable this for async support
+[build]
+rustflags = ["--cfg", "tokio_unstable"]
```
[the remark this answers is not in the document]
strohel: > Could we use [`LocalSet`](https://docs.rs/tokio/1.49.0/tokio/task/struct.LocalSet.html) in conjunction with a "current thread" runtime?

I was intrigued, but then I recalled the original reason for choosing `LocalRuntime`: it isn't an equivalent of `CurrentThreadRuntime` + `LocalSet`. `LocalSet` is more like some band-aid that has some significant drawbacks, as detailed in https://github.com/tokio-rs/tokio/issues/6739#issue-2443350742:
> However, tasks on a `LocalSet` are separate from the rest of the runtime in an uncomfortable way. For example:
>  Tasks on the `LocalSet` can `tokio::spawn` to get onto the runtime, but once you're on the runtime, you can no longer `spawn_local` to get back into the `LocalSet`.
> * From the runtime's perspective, all of the tasks in the `LocalSet` behave like a single task using `FuturesUnordered`, so various runtime options such as [`event_interval`](https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.event_interval) behave in a very surprising way by counting many tasks at once.
> * After discussions with Deno, the `LocalSet` has been shown to involve considerable performance overhead compared to unsafely spawning the `!Send` tasks onto a current-thread runtime.

It _would_ compile and work, but all downstreams would need to take care to _never_ do `tokio::spawn()` (it exits the `LocalSet`) -> `tokio::task::spawn_local()` (it panics because it is run outside `LocalSet`).

`LocalRuntime` behaves better: at any time you can use both `spawn()` and `spawn_local()` for `Send` futures and `spawn_local()` for `!Send` futures.

src/async.rs
```diff
@@ -0,0 +1,503 @@
...
+    {
+        AsyncSpawnBuilderWithoutAddress { system: self, factory: future::ready(actor) }
+    }
+
+    /// Similar to [`Self::prepare_async()`], but an async actor factory is passed instead
+    /// of an [`AsyncActor`] itself. This is used when an actor needs to be
+    /// created on its own thread instead of the calling thread.
+    /// Returns an [`AsyncSpawnBuilderWithoutAddress`] which has to be further
+    /// configured before spawning the actor.
+    pub fn prepare_async_fn<A, F>(
```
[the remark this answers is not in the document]
strohel: Hmm, maybe then `prepare_async_with_future()`? As in "prepare an async actor spawing by using a future". Or `prepare_with_future()`, but I'm kinda keen on mentioning `async` so that it rings the bell of `AsyncActor`.

`prepare_future()` sounds to me like "prepare this future", but we're preparing and actor instead.

src/async.rs
strohel: I think I saw that article before! But [sometimes the solution is worse than the problem](https://music.youtube.com/watch?v=lQsGOAkE7YY). :upside_down_face:

src/async.rs
bschwind: Oh true, I suppose we could leave it out in that case.

bschwind (approved):
LGTM++

src/async.rs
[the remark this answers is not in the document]
mbernat: Yeah, having `async` in the name is good and I'd prefer having either `factory` or `future` in the name as opposed to `fn`.

But please pick whichever of all the above options sounds the best to you; like Brian says, the type system will guide you either way.

src/async.rs
```diff
@@ -0,0 +1,504 @@
...
+//! [tokio-rs/tokio#7558](https://github.com/tokio-rs/tokio/issues/7558).
+//!
+//! With [`AsyncActor::handle()`] being an `async fn`, you gain an access to the wide async library
+//! ecosystem (currently those compatible with [`tokio`]), and you can employ concurrency (still
+//! within the single thread) when processing each message by using the various future combinators.
+//!
+//! But the incoming messages are still processed sequentially (the actor framework won't start
+//! multiple concurrent [`AsyncActor::handle()`] futures of a given actor). If you want to process
+//! the _messages_ concurrently, spawn an async task to handle the message and return from the
+//! `handle()` method immediately so new messages can arrive for processing.
```
goodhoko: I suppose you made this choice for flexibility: it's possible to handle messages concurrently by spawning tasks even then `handle()` is called sequentially, but it would be impossible to handle messages sequentially if the actor framework was calling `handle()`s concurrently, right?

src/async.rs
```diff
@@ -0,0 +1,504 @@
...
+    ) -> Result<(), ActorError>
+    where
+        F: IntoFuture<Output = A> + Send + 'static,
+        A: AsyncActor,
+    {
+        // Hold the lock until the end of the function to prevent the race
+        // condition between spawn and shutdown.
+        let system_state_lock = self.handle.system_state.read();
+        match *system_state_lock {
+            SystemState::ShuttingDown | SystemState::Stopped => {
```
goodhoko: optional nit: this would read easier for me, also lowers the indentation.

```suggestion
        if matches!(*system_state_lock, SystemState::ShuttingDown | SystemState::Stopped) {
```

examples/async_actors.rs
```diff
@@ -0,0 +1,210 @@
...
+                    "Waiting for {pending_task_count} pending tasks in Sorter before shutting \
+                     down. {} tasks already finished.",
+                    initial_task_count - pending_task_count
+                );
+
+                let results = join_all(mem::take(&mut self.pending_tasks)).await;
+                // Propagate errors from the tasks:
+                results.into_iter().try_for_each(|result| result.map_err(Error::from).flatten())?;
+
+                log::debug!("Tasks finished, shutting down the actor system.");
```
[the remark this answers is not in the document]
strohel: Yeah that could happen in theory. But I don't think it happens in practice: system_handle.shutdown() doesn't kill the actors: it sends them a message "please stop". That message has higher priority than even high-prio messages, but doesn't preempt a running handle(). It could only happen if the collector is slow and our last message is still in its inbox (because it was doing something else) when we shut down.

Maybe just a comment? This is just an example, doesn't intend to be "the best practice" in parts not directly related to async actor.

.github/workflows/clippy.yml
```diff
@@ -18,3 +18,8 @@ jobs:
         with:
           command: clippy
           args: --all-targets -- -D warnings
+      # Do another pass with all features (the `async` one for the time being)
+      - uses: actions-rs/cargo@v1
+        with:
+          command: clippy
+          args: --all-targets --all-features -- -D warnings
```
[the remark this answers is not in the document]
strohel: Because there could be clippy lints and only show up without `--all-features` (or in general with _less_ features enabled). Typically and often unconditional imports for feature-gated pieces of code. We still want to be clippy-clean even for people not enabling all feature flags.

There is even https://crates.io/crates/cargo-hack for really serious people to perform CI with all feature _combinations_ etc. (but that's for critical crates like `serde`, not us)

src/async.rs
```diff
@@ -0,0 +1,510 @@
...
+//! within the single thread) when processing each message by using the various future combinators.
+//!
+//! But the incoming messages are still processed sequentially. The actor event loop doesn't resume
+//! until the async fn handle() future resolves: it cannot start multiple concurrent
+//! [`AsyncActor::handle()`] futures also because `handle()` holds the `&mut self` reference.
+//!
+//! If you want to process the messages concurrently, spawn an async task to handle the message and
+//! return from the `handle()` method immediately so the new messages (including control messages to
+//! stop the actor) can arrive for processing.
+//!
```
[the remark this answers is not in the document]
mbernat: Yeah, that's why my suggestion was to remove that part. Not a big deal.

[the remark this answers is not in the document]
strohel: Oh I see. I've given it another read and it isn't that bad, so let me call it good enough if you will.

PR #98, opened 2026-02-02
Fix docs build on docs.rs (hopefully), add Docs CI

strohel:
We got a failure on docs.rs with 0.12, better catch this earlier: https://docs.rs/crate/tonari-actor/latest/builds/2868173

First commit adds a CI per https://github.com/dtolnay/cargo-docs-rs - but it doesn't fail.

Second commit should fix the build, but hard to tell before publishing.

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -7,7 +7,7 @@ jobs:
     name: Cargo Clippy
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v6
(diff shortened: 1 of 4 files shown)


mbernat (approved):
LGTM, thanks!

I somehow never realized tonari actor is on docs.rs, I'll make sure to use it.

strohel:
Only way to test this is to merge this and release a patch version, which I'm gonna do.

strohel:
Yay it worked: https://docs.rs/tonari-actor/latest/tonari_actor/

